PR #1003, opened 2017-11-12
Issue 970 track changes to project json

hirokiterashima:
Changes made to a project.json via the AT should now be tracked via git. So if you go to the project directory and run "git log" and "git show" commands, you should see the history of changes.

Suggested things to test:
1. creating new project: should create a new commit
2. editing existing project
    1. adding/deleting steps and components, editing notebook: should be tracked
    2. adding/deleting assets: should not be tracked


src/main/java/org/wise/portal/presentation/web/controllers/author/project/JGitUtils.java
```diff
@@ -25,53 +25,53 @@
...
-        repository.create();
-      }
-      return repository;
+  public static JSONArray getCommitHistoryJSONArray(String fullProjectDir)
+      throws IOException, GitAPIException, JSONException {
+    Iterable<RevCommit> commitHistory = JGitUtils.getCommitHistory(fullProjectDir);
+    JSONArray commitHistoryJSONArray = new JSONArray();
+    for (RevCommit commit : commitHistory) {
+      JSONObject commitHistoryJSONObject = new JSONObject();
+      ObjectId commitId = commit.getId();
```
geoffreykwan: I think declaration of this variable is unnecessary since it's only used once. We can remove it and on the next line call commitHistoryJSONObject.put("commitId", commit.getId());

This can also be done for all the other calls to commit.put().

src/main/java/org/wise/portal/presentation/web/controllers/author/project/JGitUtils.java
```diff
@@ -25,53 +25,53 @@
...
-    File gitDir = new File(localPath, ".git");
-    if (!doCreate && !gitDir.exists()) {
-      return null;
-    } else {
-      Repository repository = FileRepositoryBuilder.create(gitDir);
-      if (doCreate && !gitDir.exists()) {
-        repository.create();
-      }
-      return repository;
+  public static JSONArray getCommitHistoryJSONArray(String fullProjectDir)
```
geoffreykwan: This function takes in String fullProjectDir as a parameter where as some of the other functions in this file take in String directoryPath as a parameter. I think we should use the same parameter name for the functions if the parameters take in the same expected values.

src/main/java/org/wise/portal/presentation/web/controllers/author/project/WISE5AuthorProjectController.java
```diff
@@ -187,38 +171,8 @@ protected void registerNewProject(
...
-            }
-          }
-        } catch (JSONException e) {
-          e.printStackTrace();
-        }
-      } catch (GitAPIException e) {
-        e.printStackTrace();
-      }
-    } catch(IOException e){
+      commitChangesToProjectJSON(commitMessage, user, newProjectPath.getAbsolutePath());
```
geoffreykwan: I think this line should be commented out for now until we figure out how to commit less often.

src/main/java/org/wise/portal/presentation/web/controllers/author/project/WISE5AuthorProjectController.java
```diff
@@ -333,84 +276,47 @@ protected void saveProject(
...
-              response.getWriter().print(commitHistoryJSONArray);
-            }
-          } catch (JSONException e) {
-            e.printStackTrace();
-          }
-        }
-      } catch (GitAPIException e) {
-        e.printStackTrace();
-      }
+      commitChangesToProjectJSON(commitMessage, user, fullProjectDir);
```
geoffreykwan: I think this line should be commented out for now until we figure out how to commit less often.

hirokiterashima:
ok thx. I've implemented changes based on @geoffreykwan feedback. Ready for review again.

PR #1045, opened 2018-01-26
The Label Component can now be authored to not use the dot and line

geoffreykwan:
There is now an "Enable Dots" checkbox option in the authoring view for the Label component. Uncheck it to turn off dots and lines. Check to make sure these changes are backwards compatible with previously saved Label Component student data.


hirokiterashima:
For existing label components, the "enable dots" option is unchecked in the authoring tool. It should be checked because all existing label components use dots.

PR #1064, opened 2018-03-02
Admins can now log in as another admin. Fixes #1061.

### LINKED ISSUE #1061: Allow Admin user to switch to any user, including other admins
### Description

Admin user should be able to log in as any other user, even if that user has an admin role.

### Steps to Reproduce

1. Log in as user with admin role
2. Try to switch to another user with an admin role

**Expected behavior:**
This should be allowed and the user switch should occur.

**Actual behavior:**
Switch user doesn't occur and the user gets logged out.

hirokiterashima:
How to test:
1. log in as admin -> log in as another admin -> should succeed
2. log in as admin -> log in as non-admin -> should succeed
3. log in as researcher -> log in as another admin -> should fail
4. log in as researcher -> log in as non-admin -> should succeed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
 script:
 - mvn test
 - ./wise.sh run &
-- sleep 180
+- sleep 300
 - npm test
 - npm run test-e2e
 notifications:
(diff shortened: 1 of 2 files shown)


geoffreykwan:
All 4 tests work properly. For test 3 when a researcher tries to log in as an admin, it logs the user out and then redirects them to the login page. I don't think it should log them out but instead should keep them logged in and then display some error message such as access denied.

hirokiterashima:
I implemented test 3 change. I agree this is a better behavior.

geoffreykwan:
Looks good.

PR #1080, opened 2018-03-12
Import work as background image

geoffreykwan:
Here are all the new ways work can be imported. For all of these, the work is converted to an image and then used as a background in the second step.

Concept Map to Draw
Concept Map to Graph
Concept Map to Label
Draw to Concept Map
Draw to Graph
Draw to Label
Embedded to Concept Map
Embedded to Draw
Embedded to Label
Graph to Concept Map
Graph to Draw
Graph to Label
Label to Concept Map
Label to Draw
Label to Graph
Table to Concept Map
Table to Draw
Table to Label

The authoring tool part as well as the VLE part should be reviewed. Use connected components to author this new functionality.


hirokiterashima:
Looks great! A few comments:

**Manual testing results (possible changes necessary)**
CM (in step 1) -> Graph (in step 2): did not work consistently.
Draw (in step 1) -> CM, Graph, Label (in step 2): Graph did not import.
Graph (in step 1) -> CM, Draw, Label (in step 2): took a long time, slowed down WISE navigation afterwards.
Label (in step 1) -> CM, Draw, Graph (in step 2): Graph did not import.

Not sure how much we should look into these, but at least CM -> Graph should work more consistently. It may be a function of my slower computer (Mac OSX 2011?)

**Code Improvements (changes needed)**
1. Change var -> let/const in *Service.generateImageFromRenderedComponentState. Also remove unnecessary comments in this function.
2. htmlController: this.mode = this.$scope.mode is repeated twice in the file.

**Other thoughts (no change necessary)**
1. This could have been better separated into smaller PR? It would have made testing and reviewing easier.
2. Ideally we'd make our files shorter and we should not need to write comments like these:
```
// end of ConceptMap class
```
3. Connected Components in advanced view for a component is not easy to find. Maybe we can make this feature more prominent and encourage more authors to use it.

PR #1085, opened 2018-03-14
Fixed some minor bugs in various parts of the Authoring Tool

geoffreykwan:
This fixed 3 things. Here's what to look for.
1. Make sure adding a step as the first step in an activity doesn't break the step numbering.
2. When a student is working on a Label component, they can click the Add New Label button anytime.
3. When authoring a table component, decreasing the number of rows or columns using the text input numbers only displays a confirm dialog if the associated rows or columns have authored text values in at least one of their cells.


hirokiterashima:
functionality looks good. You just need to resolve the conflicts.

PR #1086, opened 2018-03-15
Moved actual check for password from ChangePasswordParametersValidator

### LINKED ISSUE #1072: ChangePasswordParametersValidator should defer password check to UserService method
### Description

The validator currently checks user's password. Instead, this should call a service method which will do the actual checking. This will make the code validator more robust and easier to test.

Also update ChangePasswordParametersValidatorTest to reflect this change and clean up the file as necessary.

Similar to #1070

hirokiterashima:
...to calling isPasswordCorrect method in userService. Fixes #1072. This change caused a new issue where when student tried to change their password, it threw a hibernate NonUniqueObjectException, and I fixed this by removing the userService.retrieveById call in the validator.

What to test:
1. Run mvn test and make sure that all tests pass
2. Changing password this way should work:
a. student changes their own password via account menu
b. teacher changes their own password via account menu
c. teacher changes one student's password via classroom management tool
d. teacher batch changes all students in period via classroom management tool

For 2, check that validating for wrong password works, as well as typing in correct values for all fields.

### DIFF
--- a/src/main/java/org/wise/portal/presentation/validators/ChangePasswordParametersValidator.java
+++ b/src/main/java/org/wise/portal/presentation/validators/ChangePasswordParametersValidator.java
@@ -101,15 +101,13 @@ public class ChangePasswordParametersValidator implements Validator {
       userToCheckPasswordFor = params.getUser();
     }
 
-    if (!userToCheckPasswordFor.isAdmin()) {
-      String currentPassword = params.getPasswd0();
-      if (currentPassword != null) {
-        if (!userService.isPasswordCorrect(userToCheckPasswordFor, currentPassword)) {
-          errors.rejectValue("passwd0", "presentation.validators.ChangePasswordParametersValidator.errorIncorrectCurrentPassword");
-        }
-      } else {
-        errors.rejectValue("passwd0", "presentation.validators.ChangePasswordParametersValidator.errorCurrentPasswordMissing");
+    String currentPassword = params.getPasswd0();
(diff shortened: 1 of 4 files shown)


hirokiterashima:
Hmm. I just checked and could not reproduce any of the behaviors that you mentioned. I see "incorrect password", and it won't let me change the password. Can you re-verify?

geoffreykwan:
I figured out what it was. If you're signed in as an admin, it doesn't matter what you type into the password field because in ChangePasswordParametersValidator.java on line 104

if (!userToCheckPasswordFor.isAdmin()) {

Is this if check for when admins change other people's passwords?

hirokiterashima:
I saw the problem. I removed the check for userToCheckPasswordFor.isAdmin(). Now it will check their password.

geoffreykwan:
The recent changes look good.

I discovered a different issue when a user tries to change their password multiple times on the same page.
1. Go to the profile page where you can change your password.
2. Type in your current password, for example a.
3. Type in your new password and new password again for example b.
4. Click "Save Changes".
5. Without leaving the page, try to change your password again.
6. If you enter b as your current password, it will return an "Invalid password" message.
7. The page still expects "current password" to be your previous password (a) but it should actually expect it to be (b).

hirokiterashima:
I was able to reproduce and fix the problem when teacher changed password multiple times.

geoffreykwan:
Everything works well.

PR #1088, opened 2018-03-15
Fixed a problem with graph components re-importing work in situations where they were not supposed to.

geoffreykwan:
Make sure work only gets imported into a graph component once. 
To test this
1. Create a step that generates work.
2. Create another step that imports that work into a graph component.
3. Preview the project and do work on the first step.
4. Go to the second step. It should import the work from the first step. Draw a line on the graph.
5. Go back to the first step and change the work.
6. Go back to the second step. It should not re-import the work. You should see the previous background with your previously drawn line.


hirokiterashima:
Looks good.

PR #1109, opened 2018-03-28
Authoring Tool: Fixed a problem with the startNodeId not being updated properly in rare scenarios

geoffreykwan:
Make sure the startNodeId in the project.json is set properly when steps are deleted. If all the steps in the first activity are deleted, the startNodeId should point to the first activity group id.


hirokiterashima:
looks good and behaves as described.

PR #1125, opened 2018-04-03
Authoring Tool: When an author uploads an image to be used in a field like a background, the image name is immediately populated into the field without having to click the choose button

geoffreykwan:
1. Open a project in the Authoring Tool.
2. Open or create a step that contains a component that uses an image asset. For example a draw component because it can have a background image.
3. Click the "Choose an Image" button within the authoring for the component, like for the background.
4. Upload a new image file.
5. After the image is uploaded, the file selector should automatically close and the file name should be populated into the field. You should not need to click the "Choose" button next to the image in the file selector.


hirokiterashima:
@geoffreykwan can you resolve the conflict?

src/main/webapp/wise5/authoringTool/asset/projectAssetController.es6
```diff
@@ -293,6 +293,12 @@ class ProjectAssetController {
           .then((totalUnusedFilesSize) => {
         this.setTotalUnusedFilesSize(totalUnusedFilesSize);
       });
+      if (this.nodeId != null && this.componentId != null && this.target != null) {
```
hirokiterashima: It's not clear from reading this line what this check is for. I think it'd be clearer to put this check into a function that describes what it's checking for, eg. canChooseAsset() and call it from here.

hirokiterashima:
Looks good.

PR #1147, opened 2018-04-10
Export: Fixed a problem with student response values creating multiple cells if they contained a comma

geoffreykwan:
Create open response student work in a run with this sample text below

One day, while all the animals on Mount Hood were eating together. "Mr. Fox, I was wondering, how did our home form?" asked Mr. Mouse, who had that question on his mind for a while.

Make sure the response only shows up in one cell for "Export All Work" and "Export One Workgroup Per Row".


src/main/webapp/wise5/classroomMonitor/dataExport/dataExportController.es6
```diff
@@ -618,25 +618,7 @@ class DataExportController {
...
-                // get the student data string from the component state
-                var studentDataString = componentService.getStudentDataString(componentState);
-
-                if (studentDataString != null) {
-                    // set the response
-                    row[columnNameToNumber["Response"]] = studentDataString;
-                }
-            }
-        }
+        row[columnNameToNumber["Response"]] = this.getStudentDataString(componentState);
```
hirokiterashima: This changes the code behavior a bit. Now it's always setting row[columnNameToNumber["Response"]] to a value, whereas before it set it if componentType != null and studentDataString != null. I ran through the code a bit and things seem to be ok, but let's walk through this together and make sure that nothing broke.

hirokiterashima: We discussed this together and agree that this shouldn't be an issue.

PR #1158, opened 2018-04-18
Created different endpoints for different user types. Retrieved user and then config when the site loads and when the user changes.

geoffreykwan:
1. Test that the config is retrieved for not logged in user.
2. Log in as a student and make sure student config is retrieved.
3. Refresh page and make sure the student user and config is retrieved again.
4. Log out.
5. Log in as a teacher and make sure teacher config is retrieved.
6. Refresh page and make sure the teacher user and config is retrieved again.

### DIFF
--- a/src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
+++ b/src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
@@ -1,0 +1,33 @@
+package org.wise.portal.presentation.web.controllers.teacher;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.ui.ModelMap;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Properties;
+
+/**
+ * Controller for Student REST API
+ *
+ * @author Jonathan Lim-Breitbart
+ * @author Geoffrey Kwan
+ * @author Hiroki Terashima
+ */
+@RestController
+@RequestMapping("/site/api/teacher")
+public class TeacherAPIController {
+
+  @Autowired
+  private Properties wiseProperties;
+
+  @RequestMapping(value = "/config", method = RequestMethod.GET)
+  protected String getConfig(ModelMap modelMap) throws JSONException {
+    JSONObject configJSON = new JSONObject();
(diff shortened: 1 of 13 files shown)


src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.ui.ModelMap;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Properties;
+
+/**
+ * Controller for Student REST API
```
hirokiterashima: should be "Teacher"

src/main/java/org/wise/portal/presentation/web/controllers/user/UserAPIController.java
```diff
@@ -11,6 +13,8 @@
 import org.wise.portal.domain.user.User;
 import org.wise.portal.presentation.web.controllers.ControllerUtil;
 
+import java.util.Properties;
+
 /**
  * Controller for Student REST API
```
hirokiterashima: Should be "User REST API"

src/main/webapp/site/src/app/services/user.service.ts
```diff
@@ -1,29 +1,34 @@
...
 import { of } from 'rxjs/observable/of';
-
 import { User } from '../domain/user';
 
 @Injectable()
 export class UserService {
 
   private userUrl = 'api/user/user';
-  private user: Observable<User>;
+  private user: User = null;
```
hirokiterashima: Do we need both user and user$ variable?

src/main/webapp/site/src/app/services/config.service.ts
```diff
@@ -1,27 +1,50 @@
...
 
 @Injectable()
 export class ConfigService {
 
-  private configUrl = 'api/student/config';
-  private config: Observable<Config>;
+  private userConfigUrl = 'api/user/config';
+  private studentConfigUrl = 'api/student/config';
+  private teacherConfigUrl = 'api/teacher/config';
+  private config: Config = null;
```
hirokiterashima: Do we need both config and config$ variables?

breity:
Thanks Hiroki. I made the changes you requested. I think you're right, we don't need extra config and user variables and can just use a BehaviorSubject for each. Tested and working for me. If you could go ahead and test again, that would be great.

I also made the http.get requests for user, config, and runs use the 'Cache-Control': 'no-cache' header, which prevents the responses from being cached by the browser after a user logs out.

PR #1221, opened 2018-05-24
Fixed search bug in new Site homepage library after hitting reset

breity:
Steps to test:

1. Type keyword in search box and hit enter
2. Click on reset
3. Type same keyword in search box and hit enter
4. Make sure the search works and you see the same result you did in #1


breity:
@hirokiterashima Ah, thanks for catching that. Should be fixed now.

PR #1224, opened 2018-05-29
Issue 777 concept map allow hide node titles

geoffreykwan:
Make sure the new "Show Node Labels" checkbox works properly in the Concept Map authoring view. When it is not checked, the text labels for the nodes should not show up when the student places a node on the canvas.

Previously authored Concept Map components and new Concept Map components should have "Show Node Labels" automatically set to true.


hirokiterashima:
Latest commit fixed the problem.

PR #1226, opened 2018-05-29
Fixed an error that was thrown when a teacher tries to download an export

geoffreykwan:
Try to download "Export All Work" for a run.

Before the fix, an error would be thrown and the export file would not be downloaded.
After the fix, the export file should be downloaded.


hirokiterashima:
I'm having trouble reproducing the error on the develop branch, before switching to the issue branch.

I tried "Export All Work" -> "Export All Steps" and "Select steps to export"

Issue #1344, opened 2018-06-27
Refactor animationAuthoringController.es6


geoffreykwan:
Remove unnecessary comments and clean up logic.

PR #1345, opened 2018-06-28
Refactored the animation authoring controller

geoffreykwan:
Make sure the Animation component still works in the Authoring Tool.
#1344


src/main/webapp/wise5/components/animation/animationAuthoringController.es6
```diff
@@ -38,559 +38,286 @@ class AnimationAuthoringController extends AnimationController {
...
-      }
+  authoringDeleteDataSourceClicked(animationObject) {
+    if (confirm(this.$translate('animation.areYouSureYouWantToDeleteTheDataSource'))) {
+      delete animationObject.dataSource;
+      this.authoringViewComponentChanged();
     }
+  }
 
-    // the authoring component content has changed so we will save the project
+  dataSourceNodeChanged(animationObject) {
```
hirokiterashima: These lines seems redundant? Why is it getting the nodeId from the dataSource and then setting it again? It's not clear from reading the code. Can you add comments?

src/main/webapp/wise5/components/animation/animationAuthoringController.es6
```diff
@@ -38,559 +38,286 @@ class AnimationAuthoringController extends AnimationController {
...
 
-      if (answer) {
-        // the author answered yes to delete the data points
+  canMoveDown(index, length) {
+    return index < length - 1;
+  }
 
-        // delete the data points
-        delete object.data;
+  authoringDeleteAnimationObjectClicked(index) {
```
hirokiterashima: I think we should remove the "Clicked()" calls from all ofthe *Clicked() function names, just like you did with other functions. It makes the code easier to read.

hirokiterashima:
Looks good.

Issue #1190, opened 2018-05-11
AT bug switching project via URL


hirokiterashima:
### Description

After switching the currently-authored project by replacing the project id in the url, the currently-authored project reverts to previously-authored project when using the prev/next buttons and step-chooser drop down.

**Operating System:** Mac OSX

**Browser:** Chrome

### Steps to Reproduce

1. Open project ABC in AT and open a step to author (say node 1). the URL should now look like .../author#/project/ABC/node/node1
2. Switch to author project XYZ by changing the url to .../author#/project/XYZ/node/node1 and hit 'enter'. The AT should load node 1 in project XYZ.
3. Use the prev/next buttons or the step-chooser drop down.

**Expected behavior:** AT should still editing project XYZ.

**Actual behavior:** AT goes back to editing project ABC.

PR #1378, opened 2018-08-06
Authoring Tool: Fixed a problem with the project switching back to a previous project id if the author changed the project id in the url and then interacted with the step tools

geoffreykwan:
1. Open a project directly in the Authoring Tool by using the url parameter
Example
http://localhost:8080/wise/author#/project/101

2. Change the project id directly in the url to load a different project
Example
http://localhost:8080/wise/author#/project/202

3. Use the step select drop down to go to any step in the project. Previously, the Authoring Tool would reload the previous project (example project id 101). Now it should not do that and should keep project id 202 loaded.

#1190


hirokiterashima:
Looks good.

PR #1501, opened 2018-09-27
Added favicon for site

breity:
To test:
- Load /site and make sure browser favicon updates to WISE logo


src/main/webapp/site/src/index.html
```diff
@@ -6,7 +6,7 @@
   <title>WISE: Web-based Inquiry Science Environment</title>
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <link rel="icon" type="image/x-icon" href="favicon.ico">
+  <link rel="icon" type="image/x-icon" href="favicon.ico?v=2">
```
hirokiterashima: Is ?v=2 necessary?

breity: So the reason that is there is because browsers very aggressively cache favicons. Clearing the browser cache usually doesn't have any effect. So that's why I put the query string in there. But I imagine it shouldn't be an issue for most people since they'll load the new site in the future or at different urls. So we could probably take it out.

breity:
@hirokiterashima Okay, I removed the query string from the icon href.

PR #1507, opened 2018-09-28
Added processing indicator when creating run and styled copy project dialog

### LINKED ISSUE #1492: Add processing indicator when creating a run
Add a "Processing..." or "Creating..." indicator and disable the submit button when users click the "Create" button in the create run dialog.

breity:
To test:
- Login as a teacher and view "Browse WISE Units".
- Try to create a run and make sure the "Create Run" button is disabled and the text changes to "Creating..." when clicked. Make sure creating run is successful.
- Try to copy projects and make sure that works.

Closes #1492.


breity:
@geoffreykwan Sure, I could take the text out about how copying can take some time. Locally, sometimes copying a project takes my machine 5-10 seconds. So that is a bit of a delay. But maybe the "Copying..." text is enough. Production server should be faster generally I suppose.

As for the copy button size changing when the text changes, I could make the button wider. Depending on the language though, the button text will result in different lengths. So that can complicate things. But I can up the default size a bit and if the size changes a little depending on the language, it's probably not a huge deal.

hirokiterashima:
I agree that "Copying..." text is enough and we can take it out.

breity:
Okay, I removed the copy project warning text.

I also came up with an alternate solution to changing the text on submit buttons when a request is processing.  I added a progress bar to the button itself. This way, the user sees that the action is in progress and we don't have to worry about changing the button size.

@geoffreykwan If this looks good, you can go ahead and merge. Thanks!

geoffreykwan:
The new progress bar looks pretty cool.

PR #1517, opened 2018-10-03
implemented sort by start-times for teacher runs list

vkorir:
sort teacher runs by start-times implementation


src/main/webapp/site/src/app/teacher/teacher-run-list/teacher-run-list.component.ts
```diff
@@ -88,11 +83,11 @@ export class TeacherRunListComponent implements OnInit {
     this.loaded = true;
   }
 
-  sortByRunIdDesc(a, b) {
-    if (a.id < b.id) {
-      return 1;
-    } else if (a.id > b.id) {
+  sortByStartTime(a, b) {
```
hirokiterashima: Rename to sortByStartTimeDesc. Also the logic is wrong: if a's startTime is less than b's startTime, a should be at the beginning of the array, so that it appears that the top of the page.

PR #1527, opened 2018-10-09
Issue 1515 update local user

### LINKED ISSUE #1515: Editing profile doesn't update local User in the UserService
### Description

When a user edit's their teacher or student profile info, the server updates correctly but the local User object (in UserService and the edit profile components) does not get updated. So, if the user navigates away from the `/profile/edit` route and comes back, the old profile info will be loaded.

### Steps to Reproduce

1. Log in as student or teacher and navigate to `/teacher/profile/edit` or `/student/profile/edit`.
2. Change some fields in the account info form and save.
3. Navigate to a different route and then go back to the edit profile page.

geoffreykwan:
Test the student edit profile
- Log in as a student
- Go to the Edit Profile page
- Change the language to something different like Spanish
- Click save
- Go back to the Student Home page
- Go back to the Edit Profile page
- The language should be set to Spanish

Test the teacher edit profile
- Log in as a teacher
- Go to the Edit Profile page
- Change all the fields
- Click save
- Go back to the Teacher Home page
- Go back to the Edit Profile page
- The fields should be re-populated to the values you recently set

Closes #1515

### DIFF
--- a/src/main/webapp/site/src/app/student/account/edit-profile/edit-profile.component.spec.ts
+++ b/src/main/webapp/site/src/app/student/account/edit-profile/edit-profile.component.spec.ts
@@ -63,6 +63,11 @@ describe('EditProfileComponent', () => {
     return fixture.debugElement.query(By.css('form'));
   };
 
+  const submitForm = () => {
+    const form = getForm();
+    form.triggerEventHandler('submit', null);
+  };
+
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       declarations: [ EditProfileComponent ],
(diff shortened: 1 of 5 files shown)


src/main/webapp/site/src/app/student/account/edit-profile/edit-profile.component.spec.ts
```diff
@@ -106,4 +116,13 @@ describe('EditProfileComponent', () => {
     fixture.detectChanges();
     expect(submitButton.disabled).toBe(true);
   });
+
+  it('should update the user', async() => {
+    const form = getForm();
```
hirokiterashima: Move variable declaration line to right above reference to keep variable live time short. Change for other instances also.
```
    component.editProfileFormGroup.get('language').setValue('Spanish');
    const form = getForm();
    form.triggerEventHandler('submit', null);
```

Better still, can this be made into a function like "submitForm()"?

geoffreykwan:
Ok, I created a function to submit the form.

PR #1602, opened 2018-11-07
sort run by start date and scroll to new run on creating a new run 

### LINKED ISSUE #1390: When a teacher creates a new run, sort the runs by start date and then scroll to the new run
Right now when the new run is created, it is added to the top of the page, and the page scrolls to the top. It should instead be sorted by the run start date and displayed in the right place, and scroll to the newly-added run.

vkorir:
closes #1390

- create a new run with a start date before some of the existing runs. try a few dates
- verify that runs are sorted by start date
- verify that page scrolls to the new run
- review the test


vkorir:
I'm not sure if I followed the testing guidelines. Check and let me know if I should change anything.

geoffreykwan:
Good, the test you added is in the correct format.

I noticed a minor bug though. If you change the startTime for run3 to "02-02-2010" the test fails. This is because in isRunsSortedByStartTimeDesc() you're comparing whether strings come before one another which performs an alphabetical comparison. You'll need to perform a Date comparison instead.

PR #1609, opened 2018-11-09
Fixed a constraint bug that was constraining the step the constraint was created on when it should have only been constraining steps after it

### LINKED ISSUE #1607: Constraint "Make all nodes after this not visitable" makes "this" node not visitable
Expected behavior: The step the constraint is created on should be visitable.

Actual behavior: The step the constraint is created on is not visitable which means the constraint can never be satisfied.

geoffreykwan:
2. Open or create a step that is not the first step (like the second step) and has components that can be completed
3. Create a constraint with these settings

- Action: "Make all nodes after this not visitable"
- Removal Conditional: "Any"
- Removal Criteria: "Is Completed"
- Step: (the current step)

4. Go to the Project Home
5. Preview the project
6. Try to visit the step you created the constraint on (previously you would not be able to visit it, but now you should be able to visit it)

Closes #1607


hirokiterashima:
verified fix.

PR #1637, opened 2018-11-16
Added authoring tool link to the teacher homepage tabs

### LINKED ISSUE #1623: Add link to authoring tool on teacher homepage
Remove the existing "+ Create" button and add a link to authoring tool (maybe as another tab but with appearance of a button).

breity:
To test:
- Load `/teacher` and make sure the 'Authoring Tool' link works.

Closes #1623.


hirokiterashima:
Nice improvement. @geoffreykwan and I think it the current icon implies that the AT will be opened in a new tab, so can you find a different icon- one that implies that the AT will load in the current window, replacing everything else on the page? We like the behavior of the AT NOT loading in a separate tab because it's consistent with how the CM works.

breity:
Ah, yep. I was thinking about that. I do agree that we should load the AT in the same window, since we have a link back to the teacher home in the AT account menu. How about this icon: https://material.io/tools/icons/?search=open&icon=open_in_browser&style=baseline?

hirokiterashima:
I think this is better. Let's go with it.

breity:
Okay, all set.

PR #1670, opened 2018-11-30
Fixed a problem with multiple choice radio button components with multiple correct answers being incorrect when a correct choice is submitted

### LINKED ISSUE #1626: MC: if all items "is correct", feedback says incorrect on submit
### Description

When a MC item is authored so that all choices have "is correct" checked, the feedback shows "incorrect" when the item is previewed and submitted irregardless of choice.

### Steps to Reproduce

1. Author MC item. For each choice, make sure that "is correct" checkbox is selected.
2. Preview item
3. Choose an item and hit submit.

**Expected behavior:**
You should see "correct"

**Actual behavior:** 
You see "incorrect"

geoffreykwan:
1. Create a multiple choice component that is set to single answer and has multiple correct answers
2. Preview the project
3. Select a correct answer and click submit
4. It should be marked as correct (it used to be marked as incorrect no matter what you chose)

Also run the new tests and make sure they pass.

Closes #1626


src/main/webapp/wise5/test-unit/components/multipleChoice/multipleChoiceController.spec.es6
```diff
@@ -0,0 +1,259 @@
...
+
+  it('single answer single correct answer component should show incorrect when no answer is submitted', () => {
+    loadSingleAnswerSingleCorrectAnswerComponent();
+    checkAnswer();
+    expect(multipleChoiceController.isCorrect).toBeFalsy();
+  });
+
+  it('single answer single correct answer component should show incorrect when the incorrect answer is submitted', () => {
+    loadSingleAnswerSingleCorrectAnswerComponent();
+    selectSingleAnswerChoice('y82sng5vqp');
```
hirokiterashima: For better readability, change this line to 
```selectSingleAnswerIncorrectChoice()```
which then calls
```selectSingleAnswerChoice('y82sng5vqp');```

The reason is that a reader shouldn't be expected to know if "y82sng5vqp" is correct or not.

Do this for other tests.

PR #1706, opened 2018-12-12
Issue 1693 change add label behavior

geoffreykwan:
- Make sure clicking the + now automatically creates a label on the canvas.
- Make sure the new labels appear in different locations.
- Make sure the label highlighting is more noticeable when the student clicks on a label.
- Make sure the "Label Text" input receives focus after the student clicks on a label.


hirokiterashima:
I found a bug where you when you place 3 labels on the canvas, select the third label, and delete it, the label remains. I can click it but the text input and delete buttons do not appear for it anymore. 

See attached video (inside the zip file)

[label_delete_bug.zip](https://github.com/WISE-Community/WISE/files/2682074/label_delete_bug.zip)

PR #1707, opened 2018-12-12
Issue 1467 edit tested projects

### LINKED ISSUE #1467: Implement editing projects from the new library
Allow teachers to edit projects from the new library. There should be a link in the project details drop-down menu to edit a project ONLY if the user has editing permission.

vkorir:
closes #1467 

- verify that an edit button appears only for the projects you own


geoffreykwan:
It looks like #1467 already works without any of the changes in this PR. The previous PR you made indirectly made the Edit option appear.

The code that determines whether the Edit option appears is [here](https://github.com/WISE-Community/WISE/blob/48a3626b24c616c8fc5dcf4d8b701fdeb9fd7056/src/main/webapp/site/src/app/modules/library/library-project-menu/library-project-menu.component.ts#L35).

You can close this PR since it's not needed.

PR #1712, opened 2018-12-17
Styled the edit run settings dialog

### LINKED ISSUE #1680: Style Edit Run Settings dialog
The Edit Run Settings dialog form needs styling. Should also add confirmation messages for successful changes.

We should disable changing number of students per workgroup from 1-3 to 1 if there are already workgroups with more than one student.

breity:
To test:
- Log in as a teacher.
- Select "Edit Settings" from the run menu for a variety of runs.
- Test that editing all the settings work.
- Test that invalid operations are not allowed and correct error messages appear:
-- Deleting period that already has students
-- Lowering max students per team to 1 when there are already workgroups with >1 student

Closes #1680.


src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
```diff
@@ -436,8 +437,29 @@ protected String editRunStudentsPerTeam(HttpServletRequest request,
     Run run = runService.retrieveById(runId);
     JSONObject response;
     if (run.isTeacherAssociatedToThisRun(user)) {
-      runService.setMaxWorkgroupSize(runId, maxStudentsPerTeam);
-      response = createSuccessResponse();
+      Boolean canChange = true;
```
hirokiterashima:
```suggestion
      boolean canChange = true;
```

src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
```diff
@@ -436,8 +437,29 @@ protected String editRunStudentsPerTeam(HttpServletRequest request,
     Run run = runService.retrieveById(runId);
     JSONObject response;
     if (run.isTeacherAssociatedToThisRun(user)) {
```
hirokiterashima: For better readability, move this check into a new method in RunService: 
```
boolean canChange = runService.canChangeMaxStudentsPerTeam(run.getId());
```

breity: Great, thanks. I updated with the suggested changes. I also renamed the method to 'canDecreaseMaxStudentsPerTeam' to be more specific.

src/main/java/org/wise/portal/presentation/web/controllers/teacher/TeacherAPIController.java
breity: Done.

breity:
Great, thanks! @hirokiterashima I updated with your suggested changes.

PR #1734, opened 2019-01-08
Issue 1733 add wise project version

### LINKED ISSUE #1733: Add the WISE version to the project information
When project information is requested, add the WISE version (4 or 5) to the JSON data. We use the getProjectJSON() function in ControllerUtil.java to generate the JSON. The WISE version field should be added there and should be called something like wiseVersion.

On a related note, right now TeacherAPIController.java is using its own getProjectJSON() function but it's almost exactly the same as ControllerUtil.java getProjectJSON(). Make TeacherAPIController.java call the ControllerUtil.java getProjectJSON() function.

The other places we request getProjectJSON() is in ProjectAPIController.java and StudentAPIController.java so that should automatically be fixed when you update ControllerUtil.getProjectJSON().

vkorir:
close #1733 

- verify that wiseVersion is added to the project JSON data
- verify that all other API controllers use ControllerUtil.getProjectJSON


vkorir:
fixed.

PR #1735, opened 2019-01-09
Implemented the 'Fill X Number of Rows' in a Table constraint.

### LINKED ISSUE #1731: Table component constraint: Must have at least x number of rows with at least one filled in cell
Implement this for the vle as well as the authoring.

geoffreykwan:
Here's how to test
1. Create a Table component.
2. Add a "Fill X Number of Rows" constraint on that step and make it point to the Table component.
3. Make sure the "Required Number of Filled Rows" works properly.
4. Make sure the "Table Has Header Row" works properly. For example, if the table has a header row and the constraint requires 5 filled rows, the table will actually require a total of 6 filled rows.
5. Make sure the "Require All Cells In a Row To Be Filled" works properly. If this is set to false, it means a row is considered "Filled" if at least one cell has a value.

Closes #1731


src/main/webapp/wise5/components/table/tableService.es6
```diff
@@ -260,6 +260,54 @@ class TableService extends ComponentService {
...
+    for (let r = firstStudentRow; r < rows.length; r++) {
+      const row = rows[r];
+      if (this.isRowFilled(row, requireAllCellsInARowToBeFilled)) {
+        filledRows++;
+      }
+    }
+    return filledRows >= requiredNumberOfFilledRows;
+  }
+
+  isRowFilled(row, requireAllCellsInARowToBeFilled) {
```
hirokiterashima:
```suggestion
isRowFilled(row, requireAllCellsInARowToBeFilled) {
  if (requireAllCellsInARowToBeFilled) {
    return this.isAllCellsFilledInRow(row);
  } else {
    return this.isAtLeastOneCellFilledInRow(row));
  }
}
```

geoffreykwan:
I've made the code changes and updated the help doc.

PR #1794, opened 2019-02-14
We only check if Recaptcha is required on the login page or contact page now

### LINKED ISSUE #1790: Only retrieve isRecaptchaRequired on contact form page and login pages
### Description

Right now we retrieve the isRecaptchaRequired boolean value from the server when we load the main homepage. This is not necessary and slows the down the loading of the page. Only request this boolean when loading contact form and login pages.

geoffreykwan:
- Make sure we no longer call ControllerUtil.isReCaptchaRequired() except when loading the login page or contact page.
- Make sure the Recaptcha still works when it needs to
- Make sure the login and contact pages still work when Recaptcha is not configured in wise.properties

Closes #1790

### DIFF
--- a/src/main/java/org/wise/portal/presentation/web/filters/WISEAuthenticationFailureHandler.java
+++ b/src/main/java/org/wise/portal/presentation/web/filters/WISEAuthenticationFailureHandler.java
@@ -89,16 +89,8 @@ public class WISEAuthenticationFailureHandler extends SimpleUrlAuthenticationFai
       try {
         responseJSON.put("status", "failure");
         boolean isReCaptchaRequired = ControllerUtil.isReCaptchaRequired(request);
-        if (isReCaptchaRequired) {
-          responseJSON.put("requireRecaptcha", true);
-          String contextPath = request.getContextPath();
-          response.sendRedirect(contextPath + "/login/recaptcha");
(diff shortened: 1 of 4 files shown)


src/main/java/org/wise/portal/presentation/web/filters/WISEAuthenticationFailureHandler.java
```diff
@@ -89,16 +89,8 @@ public void onAuthenticationFailure(HttpServletRequest request, HttpServletRespo
...
         boolean isReCaptchaRequired = ControllerUtil.isReCaptchaRequired(request);
-        if (isReCaptchaRequired) {
-          responseJSON.put("requireRecaptcha", true);
-          String contextPath = request.getContextPath();
-          response.sendRedirect(contextPath + "/login/recaptcha");
-        } else {
-          String contextPath = request.getContextPath();
-          response.sendRedirect(contextPath + "/login");
-        }
+        responseJSON.put("isRecaptchaRequired", isReCaptchaRequired);
```
hirokiterashima:
```suggestion
        responseJSON.put("isRecaptchaRequired", ControllerUtil.isReCaptchaRequired(request));
```
and remove line above.

PR #1795, opened 2019-02-15
Check to see if ReCaptcha is enabled when application starts up. 

### LINKED ISSUE #1791: Don't check if ReCaptcha key is valid
We don't need to check if ReCaptcha key is valid. We only need to check if we are using ReCaptcha, and to do this, we should check if private/public keys are set in wise.properties. Do this check just once, when the application starts up.

hirokiterashima:
Check that ReCaptcha still works as before. 

Test to see if ReCaptcha is enabled now looks at whether the private and public keys are set in wise.properties, and happens only once, in ControllerUtil.checkReCaptchaEnabled() method.

Closes #1791.


hirokiterashima:
@geoffreykwan made requested change, please review again.

PR #1813, opened 2019-02-25
changed all userName to username

### LINKED ISSUE #1678: Make all usage of username be all lower case
Sometimes we use userName (as two words camel cased) and sometimes we use username (as one word). To be consistent we should use all lowercase username everywhere.

Also change `userNames` to `usernames`

vkorir:
- verify that all `userName` occurrences are changed to `username`
- verify that WISE build without any errors and everything works as before

closes #1678


vkorir:
@geoffreykwan fixed! 😋

PR #1817, opened 2019-02-26
Running wise.sh with dev param now also watches for wise5 changes. 

### LINKED ISSUE #1816: Edit wise.sh to run npm run watch-all-wise5 when run with "dev" parameter
Right now we only run npm run watch-all-site when running with the "dev" parameter. Update this so we're also running watch-all-wise5.

hirokiterashima:
How to test:
1. run "wise.sh dev". This will start watching for wise5 and site changes
2. make a change to a file in site. verify that the site tests run and the site is compiled.
3. make a change to a file in wise5. verify that the wise5 tests run and the change is compiled.

Closes #1816

### DIFF
--- a/wise.sh
+++ b/wise.sh
@@ -59,7 +59,7 @@ else
 fi
 
 if [ $1 = "dev" ]; then
-  # starts npm watch-all in background, which transpiles es6 to js and watches changes to sass files
+  npm run watch-all-wise5&
   npm run watch-all-site&
 fi
 


geoffreykwan:
I found a problem after I merged. Killing the script with Ctrl+C only kills the last process

npm run watch-all-wise5&
npm run watch-all-site&

So basically after I perform Ctrl+C, "npm run watch-all-site" is stopped but "npm run watch-all-wise5" is still running. How can we kill the script and kill both processes?

hirokiterashima:
You can type "fg" to bring the "npm run watch-alll-wise5" background process to the foreground. Then you can kill it with Ctrl+C.

geoffreykwan:
Looks like it kills both processes with Ctrl+C now. I'm not sure what was happening before.

PR #1825, opened 2019-03-06
Let teachers set and edit run end dates

### LINKED ISSUE #1789: Let teachers set and edit run end dates
Right now, end date for a run is set when the teacher ends the run, even if there hasn't been any activity for a long time. The "end date" should more accurately represent the time when the run was active.

Allow teachers to choose end date when setting up a run.

We can let the teacher choose the end date and have the default option be the time of last student or teacher activity. Teachers should not be able to set the end date before the start date, start date after end date.

breity:
To test:
- Log in as a teacher
- Edit the start and end dates for some runs in your Class Schedule (via "Edit Settings" in the run menu)
- Make sure the dates update correctly and the run details update on the page
- Set up some new runs both with and without end dates, make sure they are created successfully
- Make sure you can't set the end date before the start date or start date after the end date
- Check to see that the last student login date is displayed for runs that have been used by students
- Log in as a student and make sure all students runs display correctly and can be launched

Closes #1789.


breity:
Hey @geoffreykwan:

1. Thanks for catching that. Fixed it.
2. You should see the last login time on the run listing card and also in the Edit Settings dialog (see below). This is only for runs where students have actually logged in to the run at least once.
last-student-login
last-student-login2

breity:
@geoffreykwan:
3. Okay, I can make it so that teachers can remove the end date. Just thinking ahead to when we let them use a calendar interface. Having a run without an end date doesn't make a lot of sense in that case. It would show up on the calendar indefinitely for every day after the start date. I'm wondering if we actually want to start requiring end dates?

PR #1833, opened 2019-03-10
implemented warning when users choose to edit a project with a run unit

### LINKED ISSUE #1593: Add a warning when users choose to edit a run unit
When the teacher tries to open a project to edit it and the project is associated with a run, we should show a dialog popup. The dialog popup should display a message warning them that editing a currently running unit can result in data loss.

The message should say

"Warning! You will be editing the project that is used for this run. If students have already started work on this run, this may result in undesirable effects and lost data. Are you sure you wish to proceed?"

There should be two buttons in the popup dialog.
"Cancel" on the left
"Proceed" on the right

vkorir:
1. on browser 1, log in as a Teacher and create a run from a library project
2. on browser 2, log in as a Student, add and launch the run
3. on browser 1, try editing the project you created the run with
4. on browser 1, try editing the run you created

- verify that a warning pops up saying editing might lead to undesired effects if the run has started

closes #1593


breity:
Great @vkorir. Looks good! I made some UI/text changes and will merge.

PR #1854, opened 2019-03-21
Update issue templates

hirokiterashima:
Let's adopt default GitHub templates for bug report and feature requests. I removed the OS/phone section in the bug report because for most of our cases, we do not need them.

Do you see anything else that would be good to add/remove?


termux-io:
[`df854f6242bfe03fce4e0001ab3bf9f1212b36a3](Https://heroku.com)`

PR #1858, opened 2019-03-21
Fixed a problem in the graph component that allowed students to edit graphs that should only show work and not be editable 

### LINKED ISSUE #1857: Graph connected component that shows work can still be edited
### Description

1. Create two graph steps.
2. The first graph step can be edited by the student.
3. The second graph step will show the work from the first graph step.
4. Preview the project and do work on the first graph step.
5. Go to the second graph step.

Also make sure that if the student goes to the second graph step first and then goes to the first step and does work and then goes to the second step, it properly shows the work from the first step in the second step.

**Expected behavior:**

You should not be able to add points or drag the points to a new location.

**Actual behavior:**

You can't add points but you can drag points to a new location. Students should not be able to drag the points.

geoffreykwan:
1. Create two graph steps.
2. The first graph step can be edited by the student.
3. The second graph step will show the work from the first graph step (author a connected component and set it to "Show Work").
4. Preview the project and go to the second step, it should not show any work and it should not be editable.
5. Go to the first step and add some points to the graph.
5. Go to the second step, the points from the first step should now show up in this second step.
6. Try to add a point, you should not be able to.
7. Try to drag a point, you should not be able to.

Closes #1857


src/main/webapp/wise5/components/graph/graphController.es6
```diff
@@ -4293,14 +4293,13 @@ class GraphController extends ComponentController {
                   promises.push(this.setComponentStateAsBackgroundImage(componentState));
                 }
               } else {
-                // get the trials from the component state
-                promises.push(this.getTrialsFromComponentState(nodeId, componentId, componentState));
-
                 if (type == 'showWork') {
```
hirokiterashima:
```suggestion
                if (type === 'showWork') {
```

PR #1877, opened 2019-04-02
Implemented importing discussion work

### LINKED ISSUE #1873: Support Discussion components that can be used in multiple steps.
Authors should be able to add discussion components that share the same student data across multiple steps in a project.

Right now, we can show work from another discussion component. But students can't add any new posts or comments. Curriculum authors have requested the ability to show the same full discussion (that students can continue to add to) in different sections of a unit.

hirokiterashima:
Using "connected component", the discussion component can now import from multiple other discussion components. The student can continue to add new post and comment on posts from other components.

How to test
1. As a teacher, create the following project and set up a run:
step 1: discussion component
step 2: discussion component import work from step 1
step 3: discussion component import work from step 1 and step 2 (you'll need to add 2 connected components)

2. As multiple students working at different paces (e.g. student 1 can be on step 1, student 2 can be on step 2, etc):
   1. work on step 1. It should only show posts/comments that were made on step 1 by all students in the class. Even if you go to step 3 and come back to step 1, this should still be true.
   2. go to step 2. It should show all the students' posts/comments on step 1 and you can add new posts and comment on posts from step 1 and step 2.
   3. go to step 3. It should show all the students' posts/comments on step 1 and step 2 and you can add new posts and comment on posts from step 1, step 2 and step 3.

Closes #1873


src/main/webapp/wise5/components/discussion/discussionController.es6
```diff
@@ -230,21 +263,31 @@ class DiscussionController extends ComponentController {
     });
   }
 
+  isWorkFromThisComponent(componentState) {
```
breity: Is this wrapper function necessary? Can we just call `this.isForThisComponent(componentState)` when we need it?

hirokiterashima: Good that you brought this up. I think we did this for readability ```isWorkFromThisComponent(componentState)``` 
is easier to read and understand than 
```this.isForThisComponent(componentState)```.

It is an extra method call at the cost of readability. I prefer keeping it this way. @geoffreykwan what do you think?

geoffreykwan: Yea I think isWorkFromThisComponent() makes it a little easier to read and a little more clearer to understand.

breity: Cool, makes sense.

PR #1899, opened 2019-04-29
Issue 1799 refactor graph component

### LINKED ISSUE #1799: Refactor Graph component
Refactor these files
graphAuthoringController.es6
graphController.es6
graphService.es6

Write some unit tests for these files.

geoffreykwan:
Things to test
single trial, single series
single trial, multiple series
multiple trials, each with a single series
multiple trials, each with multiple series
creating a new trial
resetting a series
undoing a change
hiding a trial
hiding all trials
deleting a trial
line graph
column graph (author the X Axis with Axis Type = "Categories" and then add some categories, add an empty data point in the series for each category in the starter data)
uneditable series
connected component show work
connected component import work
snip a graph to add it to the notebook
test student uploading a csv with graph data
importing a draw as the background to a graph
importing a label as the background to a graph
importing a concept map as the background to a graph
make sure Jenn's thermo steps with the grid and model still work
make sure the new tests pass

Closes #1799


src/main/webapp/wise5/test-unit/components/graph/graphController.spec.es6
```diff
@@ -0,0 +1,749 @@
...
+  it('should check if a series is empty', () => {
+    const series1 = [ { data: [] } ];
+    expect(graphController.isSeriesEmpty(series1)).toEqual(true);
+    const series2 = [
+      { id: 'series-0', data: [[0, 10]] }
+    ];
+    expect(graphController.isSeriesEmpty(series2)).toEqual(false);
+  });
+
+  it('should create new trial if necessary', () => {
```
hirokiterashima: Add test case where a new trial is not created

hirokiterashima:
Also remove code in graphService related to regression?

hirokiterashima:
Found one potential bug:
1. On step 1, create draw & label components
2. On step 2, create graph that imports draw as bg and another graph that imports label as bg.
3. Preview project, make draw and label on step 1, go to step 2.

I expected to see both draw and label set as bg for the two graphs on step 2, but only the second graph's bg was set.

I noticed that if you only do one of the components (draw or label but not both), the bg import works as expected.

PR #1926, opened 2019-05-29
VLE: Disabled all component inputs when a run is inactive. Also prevented client saving and server saving when a run is inactive.

### LINKED ISSUE #1912: If a run is ended, we should show an indicator/message in the student vle and also disable all inputs
### Description

1. Create a run.
2. Add a student to the run.
3. Set an end date on the run to a date in the past.
4. Students should see a message or indicator when they launch this run that has ended. We should also disable all inputs.

We should prevent student data saving on the client and server side.

geoffreykwan:
Log in as a teacher in one browser and a student in another browser.

As a teacher
1. Create a project that contains all the component types
2. Create a run with the project and set the start time to the past so the run will be active

As a student
3. Sign in as a student and add the run
4. Create student work in all the component types

As a teacher
5. Edit the run so that it is not longer active (you can set the end date to a day before today)

As a student
6. Go to the student home page
7. Launch the run in Review mode
8. Go through all the components and make sure you can't change the work and can't save anything new (this includes student work, annotations, and events)

Closes #1912


hirokiterashima:
@geoffreykwan do you have sample project with all the component types that I can use for testing?

geoffreykwan:
Here's my test project with all component types.
[83.zip](https://github.com/WISE-Community/WISE/files/3238167/83.zip)

PR #1944, opened 2019-06-19
Issue 1933 box2dmodel converter

### LINKED ISSUE #1933: Implement the converter for WISE4 box2dModel step to WISE5 Embedded component

geoffreykwan:
This PR includes code changes added to get certain functionality in the density project working.

**Test the project converter**
1. Download the density project from the WISE production library https://wise.berkeley.edu/teacher/library/tested
2. Add this line to your application.properties file and restart your local server. This is required because the converted project.json file is large and exceeds the default Tomcat 2MB post limit
```
server.tomcat.max-http-post-size=10MB
```
3. Import the WISE4 density project to your local WISE server
4. Open the WISE4 project in the Authoring Tool and convert it to WISE5
5. Make sure the project converts without error. The density models will not work because the [Box-2D-Model](https://github.com/WISE-Community/Box-2D-Model) folder needs to be added to the assets folder. There were also many other manual changes that were needed. We will test them in my converted project next.

**Test my converted density project**
1. Download the density project that I have converted and cleaned up
[144.zip](https://github.com/WISE-Community/WISE/files/3387728/144.zip)
2. Compare the WISE4 project with the WISE5 project and make sure
 - there are no missing steps
 - instructions/prompts are valid and accurate
 - density models work, save data, and reload data
 - table, graph, open response, multiple choice work is properly imported from one step to another
 - graphs work properly when they have lines and points on a single graph
 - branching works properly

Closes #1933


src/main/webapp/wise5/components/componentController.es6
```diff
@@ -647,6 +647,20 @@ class ComponentController {
     // overridden by children
   }
 
+  getConnectedComponentsAndTheirComponentStates() {
+    const connectedComponentsAndTheirComponentStates = [];
+    for (let connectedComponent of this.componentContent.connectedComponents) {
```
hirokiterashima:
```suggestion
    for (const connectedComponent of this.componentContent.connectedComponents) {
```

src/main/webapp/wise5/components/componentController.es6
```diff
@@ -647,6 +647,20 @@ class ComponentController {
...
+      const componentState = this.StudentDataService.getLatestComponentStateByNodeIdAndComponentId(
+        connectedComponent.nodeId, connectedComponent.componentId);
+      const connectedComponentsAndomponentState = {
```
hirokiterashima:
```suggestion
      const connectedComponentsAndComponentState = {
```

src/main/webapp/wise5/components/componentController.es6
```diff
@@ -647,6 +647,20 @@ class ComponentController {
...
+        connectedComponent: connectedComponent,
+        componentState: this.UtilService.makeCopyOfJSONObject(componentState)
+      };
+      connectedComponentsAndTheirComponentStates.push(connectedComponentsAndomponentState);
```
hirokiterashima:
```suggestion
      connectedComponentsAndTheirComponentStates.push(connectedComponentsAndComponentState);
```

geoffreykwan:
I've updated the project to fix the problems. Here is the updated project
[144.zip](https://github.com/WISE-Community/WISE/files/3387728/144.zip)


Step 1.2: There is no submit button. Remove "...then press "submit" to receive guidance." - **I've removed the text "Try the following, then press "submit" to receive guidance"**
Step 1.6: (probably safe to ignore) Each time "Export Model" is pressed, it says "Good. You made a single block of Gold/Brass" in W4, but not in converted project. Model feedback not working? - **I think we can leave this as is without the popup message**
Step 1.8: show previous work appears above the model - **In WISE4 the show previous work could only be displayed at the top of the step. I think it makes more sense to show the previous work closer to the prompt that asks if their predictions were correct which we can now do in WISE5.**
Step 2.5: not clear who is saying what. Make a simple styling improvement - **Fixed by moving Aida image to the left side**
Step 2.6c typo: "A block of the new material C will be lighter than and equal size block of material A…" - **Fixed by changing 'and' to 'an'**
Step 2.14 missing image of airplane and naval architects and totem pole in the show previous work. - **Fixed assets so they show up**
Step 3.2 Remove “... then press "Save" to get more guidance.” - **Removed the text**
Step 3.5 "What did you notice about the previous table?" => "What did you notice about the table?” - **Fixed**
Step 4.2 Remove “... then press "Save" to get more guidance.” - **Removed the text**
Step 4.7 "In this project you should understand these concepts" - **Fixed grammar**

PR #1946, opened 2019-06-19
Issue 1945 only include one language in site build

### LINKED ISSUE #1945: Only include one language messages.xlf file when building the site
### Description

Right now the main.js file that is built contains all the messages.*.xlf files from all locales. It should only include one messages.xlf file.

Create build configurations for all the languages we plan to support and make the build use fileReplacements to write the specific messages.*.xlf file to messages.xlf.

geoffreykwan:
1. In the git/WISE folder, run ```ng build -c=production```
2. The main.js file should be created. It used to be about 7.6 MB but now it should be about 3.6 MB because it should no longer include all of the messages.*.xlf files from all the languages.

Closes #1945


hirokiterashima:
Good points. We'll work on them in separate issues: https://github.com/WISE-Community/WISE/issues/1958 and https://github.com/WISE-Community/WISE/issues/1957

breity:
Great, LGTM.

Issue #1891, opened 2019-04-23
Use Spring Session with Redis


hirokiterashima:
### Description

Use Spring Session so we can support clustered sessions without being tied to an application container (e.g. Tomcat).

Resources:
SpringFramework Spring Session https://spring.io/projects/spring-session
Guide to Spring Session (with Redis) https://www.baeldung.com/spring-session

Make sure that these continue to work after this change:
- Users can log in
- Users can refresh the page and still be logged in
- Authors will be notified when there are other authors editing the same project
- Admin can see currently logged in users and authors

PR #1950, opened 2019-06-25
Spring Session with Redis

hirokiterashima:
Run ```npm install``` and ```maven clean compile``` to download new dependencies.

Test that everything works as before. Specifically, test that
1. You can log in with Google user and normal username/password user
2. Features that use the WebSocket (pause screens, discussion) still works


hirokiterashima:
Actually, my problem was the opposite: when signing in with username/password with a user who has a google account, it asks the user to sign-in again with with google account (shows the account chooser page). The server and UI shows the user as being signed in already. This was an edge case because users who signed up with google would not have had a password. This was happening because of this commit: 57b39ff2e9e1cab6def74ec7335eacb6b9d22dc4. We fixed this in this commit: bcec0d7203ce56c5f16efd322e3bab4f3fd2baff.

@breity perhaps you don't have the latest code, or you have local changes to code or properties file that is causing the behavior that you see? @geoffreykwan and I did see the same behavior that you described some time ago and fixed it in this commit: 3d48fad923992931632a67fc74a677f3fafe0499.  With the latest code from this branch, we were able to log in with google user both locally and on QA.

breity:
Okay, just pulled the latest, removed node_modules and re-installed, and Google sign-in is working now. Looks like this is good to go. Thanks!

breity:
Closes #1891.

PR #1967, opened 2019-07-17
Issue 1870 ProjectAPIController.copyProject() should return an error code instead of empty string

### LINKED ISSUE #1870: ProjectAPIController.copyProject() should return an error code instead of empty string
When the signed in teacher does not have permission to copy a project but makes a request to copy, we send back an empty string. We should send back an error code so that the client can then display an error message. Currently when an empty string is sent back to the client, nothing happens and the page just hangs leaving the user wondering what is happening.

geoffreykwan:
Test error message with incorrect project id
1. Sign in as a teacher
2. Go to the project library and click on a project
3. In the project popup, click the 3 dots and choose "Copy"
4. Open copy-project-dialog.component.ts in the browser and set a breakpoint on line 39
```
this.libraryService.copyProject(this.data.project.id)
```
5. Click "Copy"
6. When the breakpoint is reached, go to the console and set this.data.project.id to an id that doesn't exist
```
this.data.project.id = 10000
```
7. Click play to let the code resume
8. You should see the snack bar message that says "There was an error trying to copy the project. Please refresh the page and try again."

Test error message when logged out
1. Sign in as a teacher
2. Go to the project library and click on a project
3. In the project popup, click the 3 dots and choose "Copy"
4. Open the WISE site in second tab and sign out the user
5. Go back to the first tab and click "Copy"
6. You should see the snack bar message that says "There was an error trying to copy the project. Please refresh the page and try again."

Test that copy project works correctly when everything works properly.

Closes #1870

### DIFF
--- a/src/main/java/org/wise/portal/presentation/web/controllers/project/ProjectAPIController.java
+++ b/src/main/java/org/wise/portal/presentation/web/controllers/project/ProjectAPIController.java
@@ -21,6 +21,8 @@ import org.wise.portal.presentation.web.controllers.ControllerUtil;
 import org.wise.portal.presentation.web.controllers.author.project.WISE5AuthorProjectController;
 import org.wise.portal.service.portal.PortalService;
 import org.wise.portal.service.project.ProjectService;
+import org.wise.portal.service.user.UserService;
+import org.wise.vle.web.SecurityUtils;
 
 /**
  * Project REST API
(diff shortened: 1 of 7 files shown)


src/main/java/org/wise/portal/presentation/web/controllers/project/ProjectAPIController.java
```diff
@@ -21,6 +21,8 @@
 import org.wise.portal.presentation.web.controllers.author.project.WISE5AuthorProjectController;
```
hirokiterashima: Remove unused import.

src/main/java/org/wise/portal/presentation/web/controllers/project/ProjectAPIController.java
```diff
@@ -21,6 +21,8 @@
...
 import org.wise.portal.service.portal.PortalService;
 import org.wise.portal.service.project.ProjectService;
+import org.wise.portal.service.user.UserService;
```
hirokiterashima: Remove unused import.

src/main/java/org/wise/portal/presentation/web/controllers/project/ProjectAPIController.java
```diff
@@ -145,17 +147,16 @@ private JSONObject populateProjectMetadata(JSONObject projectLibraryGroup) throw
 
   @PostMapping("/copy")
   protected String copyProject(@RequestParam("projectId") String projectId) throws Exception {
+    JSONObject response = ControllerUtil.createFailureResponse("copyProjectError");
```
hirokiterashima: It's strange to see the error as the first line of this function, I think because the error hasn't occurred yet, and the error should happen rarely. I think it'd be better if we return the response immediately instead of storing it in a variable.

PR #1974, opened 2019-07-18
Merged wise.properties into application.properties

### LINKED ISSUE #1969: Move settings from wise.properties to application.properties
Merge all settings from wise.properties into application.properties so we don't have settings in multiple files.

- Update hibernate settings to use spring boot settings ("spring.datasource.*")
- Remove c3p0 settings. We are now using Hikari.
- Move propertyConfigurer, i18nProperties and wiseProperties from XML bean config to Java bean config

hirokiterashima:
1. Make a backup of src/main/application.properties and then remove it
2. Copy src/main/application_sample.properties to src/main/application.properties and edit the settings to match your dev environment (db, google client id/secret, etc)
3. Test that things work as before

Closes #1969


src/main/java/org/wise/Application.java
```diff
@@ -22,6 +25,27 @@ public static void main(String[] args) {
...
 
+  @Bean
+  public PropertySourcesPlaceholderConfigurer propertySourcesPlaceholderConfigurer() {
+    PropertySourcesPlaceholderConfigurer propertySourcesPlaceholderConfigurer = new PropertySourcesPlaceholderConfigurer();
+    propertySourcesPlaceholderConfigurer.setLocations(new ClassPathResource("application.properties"));
+    return propertySourcesPlaceholderConfigurer;
+  }
+
+  @Bean
+  public PropertiesFactoryBean wiseProperties() {
```
geoffreykwan: Would it be good to rename this from wiseProperties to applicationProperties for the sake of consistency and to avoid confusion? We would also need to change all the wiseProperties.get() calls to applicationProperties.get().

hirokiterashima: Good point. How about ```appProperties``` for short?

hirokiterashima: Renamed to appProperties as discussed.

PR #1996, opened 2019-08-01
Changed the contact page error message to match with the field

### LINKED ISSUE #1966: Contact WISE form shows incorrect error messages for Summary and Description fields
### Description

In the Contact WISE form which is located at https://wise.berkeley.edu/contact or http://localhost:8080/contact

 - when the Summary field is left blank, it says "City required" but it should say "Summary required"
 - when the Description field is left blank, it says "State required" but it should say "Description required"

To trigger the error messages to show up, click inside the input to give it focus and then click somewhere else on the page to make the input lose focus.

Screen Shot 2019-07-17 at 12 50 54 PM

Cipherkai:
1. Go to wise and click on the Contact WISE on the bottom of the page, then scroll to the summary field. 

2. Click inside the summary field then click out, this will make it lose focus and show the new error message. 

3. Then scroll to the description field and once more click inside then click out, making it lose focus and getting a new correct error message.

Closes #1966

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk8
+  - openjdk8
 before_install:
 - nvm install 10.16.0
 - nvm use 10.16.0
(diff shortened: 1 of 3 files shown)


Cipherkai:
@geoffreykwan i made all the changes you requested

PR #2003, opened 2019-08-05
Session Service should send a keepalive message to the server when the client detects the user is active #1995

### LINKED ISSUE #1995: Session Service should send a keepalive message to the server when the client detects the user is active
### Description

We may be able to make a request to renewSessionURL in the sessionService.es6 renewSession() function.

geoffreykwan:
Test to make sure the client sends a keepalive request to the server when the user is active in the VLE, Authoring Tool, and Classroom Monitor.

You can modify these numbers to a lower value for testing

application.properties
server.servlet.session.timeout=30m

sessionService.es6
this.showWarningInMinutesInterval = 25;
this.forceLogoutAfterWarningInMinutesInterval = 5;

Closes #1995


geoffreykwan:
Yea we should probably be using the same timeout value from application.properties. I'll work on making that change.

I think the warning should show up a fixed number of minutes before the forced logout. Maybe somewhere between 1 and 5 minutes.

src/main/webapp/wise5/services/sessionService.es6
```diff
@@ -4,14 +4,25 @@ class SessionService {
...
+    this.forceLogoutAfterWarningInterval = intervals.forceLogoutAfterWarningInterval;
+    this.checkMouseEventInterval = this.convertMinutesToMilliseconds(1);
     this.updateLastActivityTimestamp();
     this.initializeListeners();
     this.initializeSession();
   }
 
+  calculateIntervals(sessionTimeout) {
+    const forceLogoutAfterWarningInterval = 
+        Math.min(sessionTimeout * 0.1, this.convertMinutesToSeconds(5));
```
hirokiterashima: The number "5" appear out of context to anyone reading the code. It'd be better if they were set to a variable, e.g. defaultWarningInterval

PR #2019, opened 2019-09-09
NullPointerException is thrown when a student signs out from the VLE

### LINKED ISSUE #1986: NullPointerException is thrown when a student signs out from the VLE
### Description

1. Log in as a student
2. Launch a unit
3. Sign out from the VLE
4. The exception below will be thrown on the server

```
java.lang.NullPointerException
	at org.wise.vle.web.SecurityUtils.isStudent(SecurityUtils.java:205)
	at org.wise.vle.web.StudentStatusController.saveStudentStatus(StudentStatusController.java:181)
```

geoffreykwan:
1. Go to sessionService.es6 and comment out line 119. This will prevent the client from forcing a log out before the session ends on the server. We want to simulate the server session ending and then after that have the client make a request to log out.
```
//this.$rootScope.$broadcast('logOut');
```
2. Go to application.properties and set
```
server.servlet.session.timeout=1m
``` 
3. Restart the server
4. Log in as a student
5. Launch a run
6. Wait 2 minutes for the server session to end
7. Sign out of WISE by clicking on the upper right icon and then clicking "Sign Out"
8. You should no longer see a NullPointerException on the browser

Closes #1986

### DIFF
--- a/src/main/java/org/wise/portal/spring/impl/WISELogoutHandler.java
+++ b/src/main/java/org/wise/portal/spring/impl/WISELogoutHandler.java
@@ -38,14 +38,18 @@ import org.springframework.security.web.authentication.logout.LogoutHandler;
 import org.springframework.session.Session;
 import org.wise.portal.service.session.SessionService;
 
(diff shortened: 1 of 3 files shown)


hirokiterashima:
looks good!

PR #2020, opened 2019-09-09
When a user stays logged in long enough for the session to timeout, they are shown this error page

### LINKED ISSUE #2017: When a user stays logged in long enough for the session to timeout, they are shown this error page
### Description

Looks like the client session times out and then tries to redirect the user to http://localhost:8080/logout which then throws an error on the server because the user is no longer logged in.

```
Whitelabel Error Page
This application has no explicit mapping for /error, so you are seeing this as a fallback.

Fri Sep 06 13:34:46 EDT 2019
There was an unexpected error (type=Internal Server Error, status=500).
No message available
java.lang.NullPointerException
	at org.wise.portal.spring.impl.WISELogoutHandler.logout(WISELogoutHandler.java:48)
	at org.springframework.security.web.authentication.logout.CompositeLogoutHandler.logout(CompositeLogoutHandler.java:54)
	at org.springframework.security.web.authentication.logout.LogoutFilter.doFilter(LogoutFilter.java:109)
```
Screen Shot 2019-09-06 at 2 11 04 PM

geoffreykwan:
1. Go to sessionService.es6 and comment out line 119. This will prevent the client from forcing a log out before the session ends on the server. We want to simulate the server session ending and then after that have the client make a request to log out.
```
//this.$rootScope.$broadcast('logOut');
```
2. Go to application.properties and set
```
server.servlet.session.timeout=1m
``` 
3. Restart the server
4. Log in as a teacher
5. Open a project in the Authoring Tool
6. Wait 2 minutes for the server session to end
7. Sign out of WISE by clicking on the upper right icon and then clicking "Sign Out"
8. You should no longer see a NullPointerException on the browser

Closes #2017

### DIFF
--- a/src/main/java/org/wise/portal/presentation/web/controllers/author/project/WISE5AuthorProjectController.java
+++ b/src/main/java/org/wise/portal/presentation/web/controllers/author/project/WISE5AuthorProjectController.java
@@ -811,7 +811,7 @@ public class WISE5AuthorProjectController {
   private ModelAndView authorProjectEnd(@PathVariable String projectId) throws Exception {
     User user = ControllerUtil.getSignedInUser();
     Project project = projectService.getById(projectId);
-    if (projectService.canAuthorProject(project, user)) {
+    if (user != null && projectService.canAuthorProject(project, user)) {
       sessionService.removeCurrentAuthor(project.getId(), user.getUserDetails());
       notifyCurrentAuthors(projectId);
       return null;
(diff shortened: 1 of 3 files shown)


src/main/java/org/wise/portal/presentation/web/controllers/author/project/WISE5AuthorProjectController.java
```diff
@@ -811,7 +811,7 @@ private ModelAndView authorProjectBegin(@PathVariable String projectId) throws E
   private ModelAndView authorProjectEnd(@PathVariable String projectId) throws Exception {
     User user = ControllerUtil.getSignedInUser();
     Project project = projectService.getById(projectId);
-    if (projectService.canAuthorProject(project, user)) {
+    if (user != null && projectService.canAuthorProject(project, user)) {
```
hirokiterashima: Instead of checking for null user, we should prevent anonymous user access to /project/notifyAuthor* path.
In WebSecurityConfig.java, add:
```
        .antMatchers("/project/notifyAuthor*/**").hasAnyRole("ADMINISTRATOR,TEACHER")
```

geoffreykwan:
Merged into #2019.

PR #2032, opened 2019-09-18
Graph Component: Column graph tooltip does not show up on mouse over if the series is not editable

### LINKED ISSUE #2030: Graph Component: Column graph tooltip does not show up on mouse over if the series is not editable
### Description

1. Author a graph component
1. Set the Plot Type to Column Plot
1. Set the X Axis, Axis Type to Categories
1. Add two Categories "Cat" and "Dog"
1. Delete the existing series
1. Add a new series
1. Uncheck "Can Student Edit"
1. Add two data points. 50 for Cat and 25 for Dog.
1. Preview the project
1. Mouse over the bars. The tooltip should display but it doesn't.

geoffreykwan:
Test Column Plot
1. Author a graph component
1. Set the Plot Type to Column Plot
1. Set the X Axis, Axis Type to Categories
1. Add two Categories "Cat" and "Dog"
1. Delete the existing series
1. Add a new series
1. Uncheck "Can Student Edit"
1. Add two data points. 50 for Cat and 25 for Dog.
1. Preview the project
1. Mouse over the bars. The tooltip should display.

Test Line Plot
1. Author another graph component
1. Leave the Plot Type as Line Plot
1. Add an additional series. Set the name to Actual and color to red.
1. Preview the project
1. Draw two points for the Prediction series.
1. Draw two points for the Actual series.
1. Mouse over the points. The tooltip should only show up for the active series that can be currently edited. We want this behavior because showing the tooltip for all series (including the non active series) would interfere with editing the active series.

Closes #2030


src/main/webapp/wise5/components/graph/graphController.es6
```diff
@@ -658,12 +658,23 @@ class GraphController extends ComponentController {
   }
 
   setAllSeriesFields(series) {
+    const canAllSeriesMouseTrack = this.getNumberOfEditableSeries(series) == 0;
```
hirokiterashima:
```suggestion
    const canAllSeriesMouseTrack = this.getNumberOfEditableSeries(series) === 0;
```

PR #2057, opened 2019-10-16
Allow students to add hyperlinks to discussion posts and replies

### LINKED ISSUE #2053: Allow students to add hyperlinks to discussion posts and replies
When students type a url into a discussion post or reply, it should be automatically parsed as an anchor ```<a>``` tag when displayed in the discussion.

The ```<a>``` tag should have `target="_blank"` set.

geoffreykwan:
In a Discussion component, post responses and replies that contains links. All these forms of text links should be converted into clickable links.

google.com
www.google.com
http://google.com/
https://google.com/
http://www.google.com
https://www.google.com

Closes #2053


src/main/webapp/wise5/components/discussion/classResponse.es6
```diff
@@ -9,15 +9,42 @@ class ClassResponseController {
...
+        reply.studentData.responseText = this.injectLinks(reply.studentData.response);
+      }
+    }
+  }
+
+  injectLinks(response) {
+    return response.replace(this.urlMatcher, (match) => {
+      let matchUrl = match;
+      if (!match.startsWith('http')) {
+        matchUrl = '//' + match;
```
hirokiterashima: Add a comment here about this special '//[url]' notation, something like "//[url] forces the browser to figure out the correct protocol" ?

PR #2076, opened 2019-10-28
Write tests for HibernateRunDao

### LINKED ISSUE #2075: Write tests for HibernateRunDao
### Description

Test all the queries in HibernateRunDao.java.

geoffreykwan:
Run the tests in HibernateRunDaoTest.java and make sure they all pass.

Closes #2075

### DIFF
--- a/src/main/java/org/wise/portal/domain/user/User.java
+++ b/src/main/java/org/wise/portal/domain/user/User.java
@@ -28,7 +28,7 @@ import org.wise.portal.domain.authentication.MutableUserDetails;
  * @author Cynick Young
  * @author Laurel Williams
  */
-public interface User extends Persistable {
+public interface User extends Persistable, Comparable {
 
   String CURRENT_USER_SESSION_KEY = "CURRENT_USER";
 
--- a/src/main/java/org/wise/portal/domain/user/User.java
+++ b/src/main/java/org/wise/portal/domain/user/User.java
@@ -28,7 +28,7 @@ import org.wise.portal.domain.authentication.MutableUserDetails;
  * @author Cynick Young
  * @author Laurel Williams
  */
-public interface User extends Persistable, Comparable {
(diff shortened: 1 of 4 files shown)


src/main/java/org/wise/portal/domain/user/impl/UserImpl.java
```diff
@@ -46,7 +46,7 @@
 @Table(name = UserImpl.DATA_STORE_NAME)
 @Getter
 @Setter
-public class UserImpl implements User {
+public class UserImpl implements User, Comparable {
```
hirokiterashima:
```suggestion
public class UserImpl implements User
```

src/main/java/org/wise/portal/domain/user/impl/UserImpl.java
```diff
@@ -115,4 +115,9 @@ public boolean equals(Object obj) {
       return false;
     return true;
   }
+
+  @Override
+  public int compareTo(Object o) {
```
hirokiterashima:
```suggestion
  public int compareTo(User o) {
```

src/main/java/org/wise/portal/domain/user/User.java
```diff
@@ -28,7 +28,7 @@
  * @author Cynick Young
  * @author Laurel Williams
  */
-public interface User extends Persistable {
+public interface User extends Persistable, Comparable {
```
hirokiterashima:
```suggestion
public interface User extends Persistable, Comparable<User> {
```

src/main/java/org/wise/portal/domain/user/impl/UserImpl.java
```diff
@@ -115,4 +115,9 @@ public boolean equals(Object obj) {
...
+    return 0;
```
hirokiterashima:
```suggestion
    return this.getId().compareTo(o.getId());
```

PR #2130, opened 2019-11-20
OER OutsideURL OER component

### LINKED ISSUE #2125: Implement OER in OutsideURL component
### Description

 - Implement the authoring for OER's inside the OutsideURL authoring
 - Display clickable model thumbnails that will populate the url which automatically renders the preview
 - Provide a link for each model that opens a new tab and displays more information about the model
 - Use a flag (isOpenEducationalResource) to denote that the OutsideURL component is an OER component

hirokiterashima:
Outside URL component has been renamed to "Outside Resource".

Test that:
1. You can create an Outside Resource component with OER by selecting from a list of pre-selected simulations that are defined in path/to/outside_url_component/resources.json.
1. You can create an Outside Resource component without OER (just URL). This should work like Outside URL before the change.
1. The list of simulations only appears for newly-created components. (i.e. when the url has not yet been specified).
   1. Each simulation has an info link that opens in a new tab
   1. Each simulation can be previewed

Also please style the component.

Closes #2125


breity:
@geoffreykwan Good suggestions. I can make those changes.

> * [ ]  Should the credit link in the VLE link to the home page instead of the model?
How about using the info link for the credits?

hirokiterashima:
The new UI looks much better. Highlighting the model tile would be nice. I think the (px; optional) is ok.

geoffreykwan:
@breity Would highlighting the border of the selected model make it easier to see the highlighted model? The green checkmark might be easily skipped over when scrolling through the models. Maybe we could highlight the border and also keep the green checkmark?

breity:
Okay, I implemented all of the requested changes (including adding a border around the selected resource). Let me know if there's anything else you see. Thanks!

breity:
@geoffreykwan Okay, all set.

PR #2185, opened 2020-01-25
AuthorProjectController refactor and test

### LINKED ISSUE #2160: Clean and test WISE5AuthorProjectController
1. Write unit tests for important methods in WISE5AuthorProjectController
1. Clean up WISE5AuthorProjectController
   1. Remove use of ControllerUtil.getSignedInUser
   1. Modify methods to return java objects instead of JSON strings
   1. Remove unused code
   1. Replace RequestMapping with GetMapping or PostMapping

hirokiterashima:
Test that all features of the WISE5 AT work as before.

Closes #2160


src/main/webapp/wise5/services/projectAssetService.js
```diff
@@ -77,16 +66,12 @@ class ProjectAssetService {
       }).success((result, status, headers, config) => {
         // Only set the projectAssets if the result is an object.
         // Sometimes it's an error message string.
-        if (typeof result === 'object') {
-          // upload was successful.
+        if (result.status === 'error') {
+          alert(result.messageCode);
+        } else {
```
geoffreykwan: @hirokiterashima 
This messageCode should be a camel case string that gets translated on the client side. Right now the sentence is sent from the server which means it won't be translated. The sentence that is returned is generated in AuthorAPIController.java. One of the errors contains the filename which means we may need to send additional parameters back to the client like

```
{
  'status': 'error',
  'messageCode': 'fileNotAllowed,
  'param1': 'virus.exe'
}
```

We should discuss how we want to handle this.

https://github.com/WISE-Community/WISE/blob/05930b0391a9e32d0f5dd2f3dc336e6ced289305/src/main/java/org/wise/portal/presentation/web/controllers/author/project/AuthorAPIController.java#L424-L437

hirokiterashima:
@geoffreykwan please review the changes I made. The endpoint will always return an object now. What do you think?

PR #2235, opened 2020-02-25
AT subroutes: import step, project rubric

### LINKED ISSUE #2223: Move AT project rubric authoring view into sub-route
Accessible under /project/project-id/rubric

### LINKED ISSUE #2231: Move AT import step view into sub-route
It should be accessible at /project/project-id/import-step

hirokiterashima:
1. Test that AT project rubric editing works as before, but in its own route (/project/id/rubric)

You should be able to do the following:
- go to project rubric route (/project/id/rubric) by clicking on the rubric button
- edit and save the rubric using the editor, including adding assets
- go back to the project view with the back button

2. Test that AT import step feature works as before, but in its own route (/project/import-step)

You should be able to do the following:
- Choose one or more steps to import from my projects/library projects
- Cancel/go back to project view

Closes #2223
Closes #2231


hirokiterashima:
1. Importing a step has two phases, and phase 1 should be /import-step/choose-step, and phase 2 should be /import-step/choose-location. This is similar to how authors choose structures.

2. I updated the importable library projects to show the official projects.

hirokiterashima:
@geoffreykwan Good points. I implemented them all. Please re-review.

geoffreykwan:
Would it be a tiny bit more efficient if the projects were filtered and then sorted as opposed to sorted and then filtered?

hirokiterashima:
@geoffreykwan good point. I made the change to filter first, then sort. We should see a slight improvement.

PR #2253, opened 2020-03-09
Added options to milestone report graphs

### LINKED ISSUE #2250: Add options to specify score categories, title color, and bar color for milestone report graphs
Add input bindings for score range (categories) for a milestone report graph. As of now, we assume 1-5 for KI scores and 1-3 for subscores. Some c-rater items have different scales (0-2 for subscores, for example).

Also add input binding for graph title and bar colors. Bar colors are hard coded, so we should set defaults and allow authors to overwrite.

breity:
Authors can now customize title and bar color for milestone report graphs. Can also override the default possible score values for c-rater annotations (defaults are 1-5 for KI, 1-3 for other subscores).

To test:
- Graph title color:
  - In the authoring tool for a unit with a milestone report, add the `title-color` attribute to any `<milestone-report-graph>` element in a milestone report `template`. Value needs to be in the format of an rgb color value. For example: `title-color="rgb(219, 113, 28)"`.
  - Verify that the graph title matches the color you set when viewing the milestone report in the grading tool.
- Bar color:
  - In the authoring tool for a unit with a milestone report, add the `bar-color` attribute to any `<milestone-report-graph>` element in a milestone report `template`. Value needs to be in the format of an rgb color value. For example: `bar-color="rgb(219, 113, 28)"`.
  - Verify that the bars in the graph are gradations of the color you set when viewing the milestone report in the grading tool.
- Custom score categories:
  - To override the default possible scores for a c-rater subscore, add a `customScoreValues` object to a `report` entry in the `achievements` project json (in the authoring tool). Replace 'dci' in the following example with the subscore id you'd like to modify. The value should be an array of integers.
  - Make sure that the bar graph for the corresponding subscore id uses categories that match the custom score values set in the authoring and that the count for each category in the graph matches the number of students who received that score.
  - Example `achievements` json:
```
"achievements": {
  "isEnabled": true,
  "items": [
    ...
    "report": {
      "audience": [
        "teacher"
      ],
      "isEnabled": true,
      "customScoreValues": {
        "dci": [0,1,2]
       },
       "templates": [
         ...
```

Resolves #2250.


geoffreykwan:
@breity 

The graph title color and bar color works. The custom score works but I have a question about the intended behavior.

 - Suppose 3 students get a score of 0. 1 student gets a score of 2.
 - The author changes the custom score values to 1, 2, 3.
 - Should all the scores get shifted so that 3 students get 1 and 1 student gets 3? Right now it looks like in this scenario only the edge score gets shifted so the output is 3 students get 1 and 1 student gets 2.

breity:
@geoffreykwan 
Yeah, so I actually don't do any shifting of scores as of now.

The default possible scores for a subscore are [1,2,3]. Let's say that 2 students get a 2, and 3 students get a 3 for a certain subscore. If the author overrides that to [0,1,2] in the report, the only scores that should show up on the graph are the students who got a 2. All the students who got a 3 would be ignored in the calculations because their scores don't match any of the possible values for the subscore.

The main reason for this feature right now is so we can handle c-rater items that score subscores on a 0-2 scale, like the Lizards_subscores item, rather than 1-3 like most others. So I wasn't really trying to shift scores, but just allow authors to match the scores in the milestone report to the ones we get back from c-rater. Does that make sense?

We could discuss shifting scores perhaps in the future, though.

geoffreykwan:
@breity 
Is the customScoreValues for KI score handled differently than the subscores? I was referring to the KI score and the function below.

In your example wouldn't the 3 students that got a 3 be converted to get a score of 2 so there would be a total of 5 students who got a 2.

```
  adjustKIScore(scoreValue, reportSettings) {
    const teacherScore = Math.round(scoreValue);
    let min = 1;
    let max = 5;
    if (reportSettings.customScoreValues['ki']) {
      min = Math.min(...reportSettings.customScoreValues['ki']);
      max = Math.max(...reportSettings.customScoreValues['ki']);
    }
    let score = teacherScore;
    if (teacherScore > max) {
      score = max;
    }
    if (teacherScore < min) {
      score = min;
    }
    return score;
  }
```

If the subscores are not all shifted, wouldn't it be impossible to get a certain score in the report? If crater returns scores in the range of [1, 2, 3] and the researcher wants to use the scores [0, 1, 2] in the report, then it would be impossible for a student to get a 0 then right?

breity:
@geoffreykwan 
Ah yes, KI score is handled differently than the subscores. The `adjustKIScore` function is only used when we're merging the c-rater KI score and a teacher score. So it checks whether the teacher has given a score outside the available values and adjusts that score in the report to fit the allowed values. Not sure if this is the best way to deal with these cases, but that's what we settled on before. It's how we decided to handle when a teacher unknowingly alters a student KI score resulting in that student's response not appearing in the report.

> If the subscores are not all shifted, wouldn't it be impossible to get a certain score in the report? If crater returns scores in the range of [1, 2, 3] and the researcher wants to use the scores [0, 1, 2] in the report, then it would be impossible for a student to get a 0 then right?

Right, the subscores (non-KI scores) won't ever be shifted. So it would impossible to get a certain score in the report if the author overrides the values so that they don't match what c-rater returns. But that's not the purpose of this functionality. The idea here is to allow the author/researcher to override the default values that we use for subscores [1, 2, 3] in the case that the specific c-rater item returns a different set of scores. So it's just to allow us to be able to match what c-rater is returning. It wasn't meant to support authors to change or shift student scores. Although that certainly might be something we want to add in the future, as we've discussed it before.

So this option just allows us to generate a report for non-standard c-rater items (like the Lizards item that will be used in the GenEx run next week). Ideally, we should probably work with ETS so that when we input a c-rater id into the authoring tool, they would return the subscore ids and their possible values. Then we could set up the report to match those automatically. For now, this allows us to match them manually.

breity:
@hirokiterashima Thanks for catching. Can you test again?

PR #2255, opened 2020-03-10
Converted javascript in AT to typescript. 

### LINKED ISSUE #2254: AT: Convert javascript to typescript
Convert javascript files to typescript in wies5/authoringTool folder

hirokiterashima:
No new features were added. I removed some redundant comments and cleaned up the code a little bit.

Test that AT behaves the same way as before.

Closes #2254


hirokiterashima:
@breity I fixed this issue. Can you pull and try again?

I don't understand why the whitespace was causing the compiler from compiling discussionController.js the way it did. It compiled the file and outputted the problem code like this:
Screen Shot 2020-03-13 at 3 55 39 PM

It looks like it tried to transpile the file (from javascript to javascript)?

When I saved discussionController.js without changing anything, prettier removed the empty space that you can see in my latest commit, and the compiler ran again. The compiled code looks like this:
Screen Shot 2020-03-13 at 4 12 09 PM

This is the same as the actual source code, which is what we want, because the sourcecode is in javascript.

PR #2263, opened 2020-03-13
Fixed issue not being able to save notebook item with MySQL8 

### LINKED ISSUE #2262: MySQL8: cannot save notebook item
### Description

Can't save a notebook item when using MySQL8.

### Steps to Reproduce

1. Try to save a new notebook item as a student

**Expected behavior:**
A notebook item should be created

**Actual behavior:**
An error is thrown on the server-side and the notebook item is not created.
```
Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'groups, localNotebookItemId, nodeId, parentNotebookItemId, periodId, runId, serv' at line 1
```

### Additional Information
This is because 'groups' is a keyword starting in MySQL8.0.2.

hirokiterashima:
Make sure that you can save a notebook item running WISE with MySQL8

Resolves #2262

### DIFF
--- a/src/main/java/org/wise/vle/domain/work/NotebookItem.java
+++ b/src/main/java/org/wise/vle/domain/work/NotebookItem.java
@@ -88,7 +88,7 @@ public class NotebookItem extends PersistableDomain {
   @Column(name = "parentNotebookItemId")
   private Integer parentNotebookItemId;
 
-  @Column(name = "groups")
+  @Column(name = "`groups`")
   private String groups;
 
   @Column(name = "type", length = 30, nullable = true)


geoffreykwan:
So the database connection worked for me on Friday but today on Monday it doesn't work anymore, maybe because I restarted my computer. I received this exception when starting up WISE

"com.mysql.jdbc.exceptions.jdbc4.MySQLNonTransientConnectionException: Could not create connection to database server. Attempted reconnect 3 times. Giving up."

So I tried updating mysql-connector-java in pom.xml to

```
    <dependency>
      <groupId>mysql</groupId>
      <artifactId>mysql-connector-java</artifactId>
      <version>8.0.19</version>
    </dependency>
```

and switching from com.mysql.jdbc.Driver to com.mysql.cj.jdbc.Driver in application.properties

Then I started receiving an error saying

"The server time zone value 'EDT' is unrecognized or represents more than one time zone. You must configure either the server or JDBC driver (via the serverTimezone configuration property) to use a more specifc time zone value if you want to utilize time zone support.)"

So I added parameters for useJDBCCompliantTimezoneShift, useLegacyDatetimeCode, and serverTimezone in the connection string. I don't think using these parameters is an ideal solution given the comments I have read about them. We may need to figure out something else.

My application.properties now looks like this
```
spring.datasource.url=jdbc:mysql://localhost:3306/wise_database?autoReconnect=true&useUnicode=true&characterEncoding=UTF-8&allowPublicKeyRetrieval=true&useSSL=false&driver=com.mysql.cj.jdbc.Driver&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC
spring.datasource.driver-class-name=com.mysql.cj.jdbc.Driver
```

hirokiterashima:
@geoffreykwan 

[mysql-connector-java 5.1.48 says it supports MySQL 8](https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1-48.html), so I would try to figure out why it stopped working before upgrading to mysql-connector-java 8. Was MySQL running when you started WISE? Did you change the username/password?

As you saw and as written [here](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-upgrading-to-8.0.html), using mysql-connector-java 8 will require changes. I think we should stick with 5.1.48 and make upgrading to mysql-connector-java 8 into another issue.

hirokiterashima:
@geoffreykwan Oops my bad. I forgot that I needed to add 
```allowPublicKeyRetrieval=true``` to the JDBC url when I upgraded to MySQL8 (while keeping everything else the same). If you look further down in the original stacktrace, I think it mentions it. 

What's strange is that once you add the allowPublicKeyRetrieval=true and the server starts up, you can remove it going forward.

PR #2273, opened 2020-03-23
Allow connected component to clear graph

hirokiterashima:
Test that a component that is connected to a graph component can clear the graph by setting the "clearGraph=true" field in the student data.

See [Snap! Thermo integration demo project](https://wise-qa.berkeley.edu/project/106#!/project/106/node5) for authoring example. The graph should have connected component authored with the "clearGraph" read field.
```
"fields": [
                {
                    "name": "clearGraph",
                    "action": "read",
                    "when": "always"
                }
            ]
```


geoffreykwan:
Should we check to make sure that ```studentData.clearGraph === true```?

hirokiterashima:
@geoffreykwan good catch. I added the check.

PR #2306, opened 2020-04-20
Refactored student data service

geoffreykwan:
Make sure saving and re-populating student data still works.


hirokiterashima:
1. To reduce the complexity of evaluateCriteria, try using a map of functionName => function as mentioned in this SO thread:
https://stackoverflow.com/questions/10175805/how-to-avoid-a-lot-of-if-else-conditions


2. change 
```
  saveToServer(componentStates, events, annotations) {
```

to have default values
```
  saveToServer(componentStates = [], events = [], annotations = []) {
```
and then you can remove the null checks.


3. change occurrences of
```
      action === 'a' ||
      action === 'b' ||
      action === 'c'
```
to
```
['a', 'b', 'c'].includes(action)
```

src/main/webapp/wise5/vle/node/nodeController.ts
```diff
@@ -674,19 +674,11 @@ class NodeController {
...
-        for (const componentState of componentStates) {
-          if (componentState != null) {
-            let annotations = componentState.annotations;
-            if (annotations != null) {
-              componentAnnotations = componentAnnotations.concat(annotations);
-            }
-            delete componentState.annotations;
-          }
-        }
+    return this.createComponentStates(isAutoSave, componentId, isSubmit)
```
hirokiterashima, line 677: Can we make nodeController.createComponentStates return an array of non-null elements? It looks like this is the crux of the problem.

PR #2348, opened 2020-05-12
Notebook grading redesign

### LINKED ISSUE #2318: Style Notebooks section of Classroom Monitor
- Style student note and report viewing
- Hide student reports if not enabled in run

breity:
Redesigned the notebook grading view in the Classroom Monitor to show a sortable list of workgroups (similar to node grading view). Notebook work (notes and/or reports, if enabled) can be toggled opened and closed for each workgroup.

To test:
- Open teacher tools for a run with student notebook enabled.
- Navigate to Notebook view and ensure that viewing student notebooks (both the notes and report tabs) works properly.
- Enable and disable notes and reports in the run. Make sure that list column headers and the work displays for each are shown when enabled and not shown when disabled.

A few related issues:
1. I've hidden the ability to score and send comments for notebook reports since this functionality is currently broken. Creating a separate issue to address.
2. Notebook work doesn't update if students change their work while the teacher tools are open. You have to reload the teacher tools to see any new notebook content. Also creating a separate issue for this.

Closes #2318.


breity:
@hirokiterashima @geoffreykwan Thanks for finding the issues. Should be fixed now except for the first two that Hiroki found. These only happen when switching from AT notebook page back to CM. We've been debugging and haven't been able to find a solution yet, so will create a new issue for that.

Requesting reviews again here from you guys for the other fixes.

geoffreykwan:
If you mouse over the step title in a note, it will become underlined and your cursor will change to the hand which may give the impression that it's clickable. Maybe we should either disable the underline and hand or make the link actually work and send the teacher to the grading view for that step and student.

breity:
@geoffreykwan Yeah I just disabled the link following for now. I thought it might be disjointing for the teacher to navigate to the step grading view from the notes and lose all the notebook context. But I'm not sure.

What do you guys think? Remove the clickable link and just display the source step text for now?

PR #2358, opened 2020-05-15
Refactored NotificationController.getNotifications(). 

### LINKED ISSUE #2357: Refactor NotificationController.getNotifications
### Description

Refactor this method to make it run faster, allow gzip compression, and easier to read.

- Return List<Notification> instead of writing string to response. This will allow gzip compression by Nginx
- Write tests

hirokiterashima:
This PR includes these changes:
- the getNotifications() method now returns ```List<Notification>``` instead of writing a JSON string to the output string, so it can be gzipped by nginx.
- removed need for preview user to retrieve notifications
- removed need for teachers to retrieve notifications that are not addressed to them. Before, we were retrieving all notifications for the run
- made Notification's to/fromWorkgroup and type fields non-nullable
- fixed issue where a new notification’s location was not displaying correctly. It was showing the same location as the last-retrieved notification (notificationMenu.html)

Test that notifications work for both teachers and students, and preview users do not retrieve notifications.

Resolves #2357


hirokiterashima:
We currently don’t send notifications addressed to shared teachers for a run. It's always addressed to the the run owner.

I think we have three options:
1. retrieve all notifications for a run for shared teachers (or retrieve for all teachers to keep it consistent?)
2. make teacher notifications addressed to shared teachers also
3. not show notifications for shared teachers unless it's addressed to them

Seems like #2 complicates things more than necessary. I don't know about #3. How often do shared teachers view and act on run notifications?

Should we stick with retrieving notifications for all teachers?

geoffreykwan:
Ideally I think all teachers should be able to see all teacher notifications. Could we retrieve all the teacher notifications when any teacher loads the Classroom Monitor?

hirokiterashima:
Yupp, made the change so we retrieve all the teacher notifications when any teacher loads the CM.

geoffreykwan:
Right now when a teacher loads the Classroom Monitor, it retrieves all the notifications for the run (even the student to student notifications). Would it be better to only retrieve the notifications from and to teachers?

PR #2361, opened 2020-05-18
Resize images before uploading as student assets

hirokiterashima:
Test that images bigger than 1000x1000 are resized and reduced in quality before being uploaded onto the server.

When testing locally, you can see the uploaded images in src/main/webapp/studentuploads directory.


hirokiterashima:
Thanks for the catch. 

The original change was for assets uploaded as attachments to OR components. I added image resizing and compression to assets uploaded to student notebooks. I also removed the centerCrop option so the images retain their height/width ratios when they're resized.

PR #2366, opened 2020-05-20
Change workgroup's period without losing work

### LINKED ISSUE #2288: Move workgroup to another period without losing work
Allow teachers to move workgroups (and all students in the workgroup) to a different period without losing work.

hirokiterashima:
With this change, in the "Manage Students" page, teams will have a "period" option to change its period as well as all of its members' periods. No new workgroup will be created, so the students can keep their work.

As before, unassigned students without a team will have a "period" option to change their period.

Test that both cases above work.

Resolves #2288


hirokiterashima:
Good catch. This was a head-scratcher and took me a long time to figure out. I finally fixed it by focusing on the submit button when the page loads.

PR #2369, opened 2020-05-22
Added Google Analytics pageview tracking on route changes

### LINKED ISSUE #2368: Re-enable Google Analytics pageview tracking
Add pageview tracking to WISE. Currently, Google Analytics is only running on legacy/admin pages.

breity:
Any change in route will now generate a pageview on Google Analytics.

To test:
- Traffic from `localhost` is being filtered in GA, so testing locally won't work without modifying Analytics settings. To do so, remove the 'localhost' filter (Admin -> Filters for the 'wise.berkeley.edu' view).
- Navigate around the site and verify that pageviews are being recorded. This can be done by monitoring real-time traffic in GA (Realtime -> Overview).
- After testing, re-enable the localhost filter by adding a new filter with these settings:
Screen Shot 2020-05-22 at 9 55 14 AM

Closes #2368.

### DIFF
--- a/src/main/java/org/wise/portal/presentation/web/controllers/user/UserAPIController.java
+++ b/src/main/java/org/wise/portal/presentation/web/controllers/user/UserAPIController.java
@@ -126,6 +126,7 @@ public class UserAPIController {
     String contextPath = request.getContextPath();
     config.put("contextPath", contextPath);
     config.put("currentTime", System.currentTimeMillis());
+    config.put("googleAnalyticsId", appProperties.get("google_analytics_id"));
     config.put("googleClientId", googleClientId);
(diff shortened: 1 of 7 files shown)


breity:
@hirokiterashima and @geoffreykwan I updated to use application.properties. Tracking should only be enabled now if `google_analytics_id` is set.

I also set tracking to only work in production environments and removed the `localhost` filter in GA. To test locally, comment out the check for production in `app.component.ts`: https://github.com/WISE-Community/WISE/blob/7dc7dd4dcb5d0a646e802e4ded37368ceae8aa36/src/main/webapp/site/src/app/app.component.ts#L100.

PR #2377, opened 2020-05-27
Converted UtilService to Angular

hirokiterashima:
Run ```npm install``` to get the updates and test that everything works as before.

I was also able to move the util service unit tests to Angular 8. They now run as part of ```npm test```.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ script:
   - mvn test jacoco:report
   - npm install
   - npm test
-  - npm run test-site
 after_script:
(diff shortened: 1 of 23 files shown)


hirokiterashima:
Ah, I was too zealous with the conversion and tried to refactor makeCopyOfJSONObject method when I shouldn't have. Reverted the code, but also saw that the null-check in the original code was not necessary because ```angular.fromJson(angular.toJson(null))``` returns null anyway, so I removed the null-check.

PR #2379, opened 2020-05-28
Handle extra node exited event in export step visits

### LINKED ISSUE #2376: Handle extra node exited event in export step visits
### Description

The student VLE has been generating an extra nodeExited event when the student enters a step causing the Visit Duration value to be 0 when it shouldn't be. Make the step visits export compensate for this.

Also, show visits for steps that were deleted from the project.

geoffreykwan:
1. Make sure extra nodeExited events are ignored. You can look at the "Export Events" to view all the individual nodeEntered and nodeExited events to see when those occur and compare it with the "Export Step Visit Events".
1. Make sure step visits to deleted steps show up in the export
1. Make sure visits without an "Exit Time" now shows "(Unknown Exit Time)" and "(Unknown Visit Duration)"
1. Make sure the "Previous Node ID" is empty for the first visit for all workgroups in the run
1. Make sure workgroups that are empty do now show up in the export. Should we show the visits for empty workgroups?

Closes #2376

### DIFF
--- a/src/main/webapp/wise5/classroomMonitor/dataExport/exportVisits.html
+++ b/src/main/webapp/wise5/classroomMonitor/dataExport/exportVisits.html
@@ -73,12 +73,18 @@
                 </md-tooltip>
               </md-button>
             </div>
-           <div ng-repeat="(key, value) in exportVisitsController.nodes"
+            <div ng-repeat="(key, value) in exportVisitsController.nodes"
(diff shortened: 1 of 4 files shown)


hirokiterashima:
I think it's ok to omit empty workgroups because the students presumably went on to join other workgroups, which we have data for.

PR #2393, opened 2020-06-04
Upgraded to angular 9.1

### LINKED ISSUE #2305: Upgrade to Angular 9.1
Upgrade to 9.1 and take advantage of the new ivy compiler.

https://blog.angular.io/version-9-1-of-angular-now-available-typescript-3-8-faster-builds-and-more-eb292f989428

hirokiterashima:
This change upgrades Angular from 8 to 9.1 and by default uses ivy and AoT compiler for development.

**To test**
Run ```npm install``` to get the updates before testing. If you have compile issues, try completely deleting node_modules folder and running ```npm install```

Test both dev (```npm run build-dev```) and prod (```npm run build-prod```) builds and make sure that everything works as before.

@breity somehow I broke the tooltip background (in run settings dialog for creating and editing a run), and it no longer shows up. I couldn't fix it. According to an article, AngularMaterial was also updated in 9.1 (https://blog.angular.io/version-9-of-angular-now-available-project-ivy-has-arrived-23c97b63cfa3), which might have caused this change. Can you please take a look?
Screen Shot 2020-06-04 at 8 30 18 AM

Resolves #2305


hirokiterashima:
Thanks for catching the issues.

It looks like bug #1 was always there. I can see it in production. We can resolve it in a separate issue.

For the other bugs, they were related to instances where we had self-closing textarea elements, like ```<textarea />```. I fixed it by closing it like this ```<textarea></textarea>```

PR #2399, opened 2020-06-08
Upgraded [vle/cm/at]ProjectService to Angular

### LINKED ISSUE #2397: ProjectService: Upgrade from AngularJS -> Angular

See #2367 and #2389 for reference

hirokiterashima:
Test that everything works as before.

One change to note (and test) is ProjectService.retrieveProject. To ensure that the function always retrieves project.json, I added "Cache-Control: no-cache" to the request header and removed the param in the GET request. This made the function easier to test.

I also moved the [vle/cm/at]ProjectService unit tests to Angular. They now run as part of ```npm test```.

Resolves #2397


hirokiterashima:
Thanks for catching the issues. I've fixed them and also fixed another issue where tooltip was not appearing above the progress bar in CM grade-by-step.

PR #2430, opened 2020-06-29
Added support for responsive images in HeroSection components

### LINKED ISSUE #2429: Optimize loading hero images
HeroSection components only accept one image src. Support responsive images to optimize loading.

breity:
HeroSection components now support multiple background images for responsive, optimized loading.

To test, load homepage at different viewport widths and verify that different size background images are downloaded:
- Viewports under 600px wide should load `wise-students-hero-600w.webp`.
- 600px to 899px viewports should load `wise-students-hero-900w.webp`.
- 900px to 1199px viewports should load `wise-students-hero-1200w.webp`.
- 1200px and higher viewports should load `wise-students-hero.webp`.
- Test in a browser that doesn't support .webp files (Safari) and verify that corresponding .jpg images are loaded in the above conditions.

Closes #2429.


hirokiterashima:
On Chrome, every time I open or refresh the homepage, it seems to be retrieving wise-students-hero*.webp twice.
On Safari, .webp files are still requested, as well as the jpg files.
Firefox seems to work fine.

Can we get around these behaviors, or are they unavoidable?

breity:
> On Chrome, every time I open or refresh the homepage, it seems to be retrieving wise-students-hero*.webp twice.
> On Safari, .webp files are still requested, as well as the jpg files.
> Firefox seems to work fine.
> 
> Can we get around these behaviors, or are they unavoidable?

@hirokiterashima Yes, it's retrieving wise-students-hero*.webp twice because I use a hidden `<picture>` element which determines which image to use for the background-image. Once the image loads the first time, the background-image src will be set and the browser will request the file again. The file will generally be cached by the browser, though, so it really should be just 1 request in practice.

Not sure about Safari. Haven't tried this on a Mac. Looks like Safari 14 added support for .webp files (https://www.zdnet.com/article/safari-14-removes-flash-gets-support-for-breach-alerts-http3-and-webp/). Does Safari download both the .jpg and .webp file for you?

PR #2459, opened 2020-07-13
Upgraded StudentStatusService to Angular

### LINKED ISSUE #2458: StudentStatusService: Upgrade from AngularJS -> Angular
- See #2367 and #2389 for reference

hirokiterashima:
Removed StudentStatusService from student files, as it was not being used.

Test that retrieving student status from teacher tool works as before.

Resolves #2458


hirokiterashima:
This has been rolled into #2461.

PR #2488, opened 2020-07-27
Upgraded NotificationService to Angular. 

### LINKED ISSUE #2472: NotificationService: Upgrade from AngularJS -> Angular
Description

See #2367 and #2389 for reference

hirokiterashima:
Also:
- added serializer and deserializer for Notification. 
- extended message converters in WebConfig.java so we no longer need @JsonSerialize/@JsonDeserialize annotations on java domain objects.

Test that creating and dismissing notifications work as before. Notifications are created when:
- teacher provides feedback to student work
- student comments on another discussion post
- cRater item is authored with enable notifications on. This creates two notifications (one for teacher, and one for student), which can both be dismissed by either the teacher or the student at the same time.

Resolves #2472


hirokiterashima:
1. I put it back in for now for consistency with other fields, but the null check is actually not necessary for componentType and other fields. We should refactor it later. The JSON response that is sent to the client becomes {“id”:4,…”componentType”: null}, which still works. I think in the future we should send null-value fields and let the client handle it, instead of omitting the field in the JSON response.

2-4 fixed! Thanks for catching them. Please test that Ambient notifications can be dismissed by both student and teacher.

geoffreykwan:
1. The function vleController.hasNewAmbientNotifications() gets called a lot when you just move the mouse around. This could probably be made more efficient but we could do this in another issue.

2. Would changing the "notificationMessage = " lines make it easier to read?

https://github.com/WISE-Community/WISE/blob/27df82b14c1043d2ba9e10bdeca8041680e6e152/src/main/webapp/wise5/services/notificationService.ts#L117-L129

We could set notificationMessage to a new variable so that we don't overwrite the notificationMessage input parameter and we could rename the input parameter to notificationMessageTemplate and then use const notificationMessage for the variable that contains all replaced values.

```
    const notificationMessage = notificationMessage
        .replace('{{username}}', this.ConfigService.getUsernameByWorkgroupId(fromWorkgroupId))
        .replace('{{score}}', notificationForScore.score)
        .replace('{{dismissCode}}', notificationForScore.dismissCode);
```

3. The teacher can dismiss the ambient notification by clicking the notification in the Classroom Monitor (you don't even need to click on the x, you can just click on the notification text). I was not expecting this behavior. I was expecting to need to click the x or to have to go to the student and enter this dismiss code. This could be addressed in another issue if we want to change this behavior.

hirokiterashima:
I created issues for 1 (#2495) and 3 (#2496). It'd be good to address these in the future. We should discuss #2496 about expected behavior.

I also cleaned up sendNotificationToUser. I didn't see a need to introduce a new variable here since notificationMessage is a string and it is passed by value. Chaining the replace() calls makes the code easier to read.

PR #2497, opened 2020-07-31
Upgraded NotebookService from AngularJS to Angular

### LINKED ISSUE #2479: NotebookService: Upgrade from AngularJS to Angular
- Upgrade to Angular, then downgrade it using downgradeInjectable so it can continue to be used in student/preview and teacher apps
- Write unit tests to run with other Angular tests

breity:
Upgraded NotebookService to Angular and wrote tests for NotebookService.

To test:
- Make sure Notebook tools all work as before:
  - Student notebook and report in vle (preview + runs)
  - Teacher notebook and notebook grading view in classroom monitor
  - Notebook authoring

Closes #2479.


breity:
Thanks @geoffreykwan. I made the requested changes.

geoffreykwan:
The spread operator performs a shallow copy so note.content and noteCopy.content are actually pointing to the same object. I think this might lead to slightly unintended behavior in certain places in the code such as in deleteNote() and reviveNote(). In this case it doesn't really matter because noteCopy will have its clientSaveTime overwritten when it gets saved shortly after that.

Now that I think about it, does note.content.clientSaveTime actually need to be updated at all if we are saving note revisions?

https://github.com/WISE-Community/WISE/blob/2d62f7098111991281d8c36dac4d0755609eb552/src/main/webapp/wise5/services/notebookService.ts#L87-L101

breity:
> Now that I think about it, does note.content.clientSaveTime actually need to be updated at all if we are saving note revisions?
> 
> https://github.com/WISE-Community/WISE/blob/2d62f7098111991281d8c36dac4d0755609eb552/src/main/webapp/wise5/services/notebookService.ts#L87-L101

Yeah, it's a little confusing. We seem to store the client save time in two places: `note.clientSaveTime` and `note.content.clientSaveTime`.  I don't see anywhere in the code where we actually use `note.content.clientSaveTime` for anything other than saving a new note or a note version. And there we just use it as a parameter for `NotebookService.saveNotebookItem()`. So maybe we don't need to be setting it at all?

PR #2498, opened 2020-07-31
Upgraded spring boot to 2.3.2

### LINKED ISSUE #2453: Upgrade to Spring-Boot 2.3.2
https://spring.io/blog/2020/07/24/spring-boot-2-3-2-available-now

We want to take advantage of the lazy-initialization when starting up. There was a bug with WebSocket that was fixed in this release.

hirokiterashima:
Clear the Redis database by running
$ redis-cli
> flushall

and test that WISE starts up and works as before.

Resolves #2453

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -332,7 +332,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.2.0.RELEASE</version>
+    <version>2.3.2.RELEASE</version>
     <relativePath/>
   </parent>
   <dependencies>


hirokiterashima:
Thanks for catching this. It looks like the issue with serialization (non) compatibility across different versions is not a bug, but a deliberate policy https://github.com/spring-projects/spring-security/issues/1945

The StringRedisSerializer fix does not work for me. I still get a 500 page with the error
```
Cannot deserialize; nested exception is org.springframework.core.serializer.support.SerializationFailedException: Failed to deserialize payload. Is the byte array a result of corresponding serialization for DefaultDeserializer?; nested exception is java.io.InvalidClassException: org.springframework.security.core.context.SecurityContextImpl; local class incompatible: stream classdesc serialVersionUID = 520, local class serialVersionUID = 530
```

There is a work-around that is mentioned in the blog https://sadique.io/blog/2016/11/02/handling-deserialization-errors-in-spring-redis-sessions/, which deletes old sessions. So the solutions are either to flush all sessions, or implementing the work-around. We discussed and agreed on flushing the sessions.

PR #2550, opened 2020-08-24
Discussion gating authoring option

hirokiterashima:
This commit allows authors to allow/disallow students to seeing other students' posts before creating new post in discussion component. The JSON field is called ```gateClassmateResponses```.

Test that you can 
- enable/disable the ```gateClassmateResponses``` option in the discussion component authoring
- see others' posts without creating a new post if gating is disabled (```gateClassmateResponses=false```)
- NOT see others' posts without creating a new post if gating is enabled (```gateClassmateResponses=true```)


hirokiterashima:
Replaced by #2553, which doesn't have any of the merge/revert commits.

PR #2552, opened 2020-08-25
Events to observables: refreshProject, scrollToBottom, toggleComponentAdvancedView

### LINKED ISSUE #2551: Replace 'refreshProject', 'scrollToBottom' event broadcasts with Observables

hirokiterashima:
These are all events in the authoring tool.

Test that:
- refreshProject() is invoked when changes are made (importing nodes in structured authoring, auto-saving node JSON, etc)
- the view scrolls to the bottom when a new constraint is added to a node
- advanced component authoring view can be toggled

Resolves #2551


hirokiterashima:
Replaced by #2570. Team agreed to implement toggleComponentAdvancedView as an input to the component instead of as an event.

PR #2594, opened 2020-09-02
Removed events: componentCompleted, nodeSaveClicked, showRevisions

### LINKED ISSUE #2588: Remove events: "componentCompleted", "nodeSaveClicked" and "showRevisions"
These events are no longer being used.

hirokiterashima:
None of these events were being used, so there should not be any change to the behavior.

Test that functionality for completion, node saving, and showing revisions in the CM still work like before.

Closes #2588


geoffreykwan (approved):
Looks good.

PR #2597, opened 2020-09-03
Converted componentStudentDataChanged event to observable

### LINKED ISSUE #2596: Replace 'componentStudentDataChanged' event broadcasts with Observables

hirokiterashima:
Also cleaned up subscribe callback. 

Test that student work for components work as before.
Test connected components and animation component in particular.

Closes #2596

### DIFF
--- a/src/main/webapp/wise5/components/animation/animationController.ts
+++ b/src/main/webapp/wise5/components/animation/animationController.ts
@@ -966,19 +966,10 @@ class AnimationController extends ComponentController {
   studentDataChanged() {
     this.setIsDirty(true);
     this.emitComponentDirty(true);
-
     this.setIsSubmit(true);
     this.emitComponentSubmitDirty(true);
-
     this.clearSaveText();
-
-    this.createComponentState('change').then(componentState => {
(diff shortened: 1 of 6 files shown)


geoffreykwan (changes requested):
Functionality works well. I found two places where we could clean up the code a little.

Instead of calling this.createComponentState(action).then() I think we can call this.createComponentStateAndBroadcast(action) which is a parent function in the componentController.ts.

https://github.com/WISE-Community/WISE/blob/77bb7024e93e13eeea1b6e821ca775a29df54c96/src/main/webapp/wise5/components/animation/animationController.ts#L966-L982

https://github.com/WISE-Community/WISE/blob/77bb7024e93e13eeea1b6e821ca775a29df54c96/src/main/webapp/wise5/components/discussion/discussionController.ts#L433-L443

hirokiterashima:
Thanks for catching it. I replaced the code with a call to componentController's createComponentStateAndBroadcast() function.

geoffreykwan (approved):
Looks good.

PR #2601, opened 2020-09-04
Extracted AT import component feature

### LINKED ISSUE #2598: Refactor: extract import component feature into its own route
Right now the import component feature in the authoring tool is embedded in nodeAuthoringController.ts and node.html and turned on/off in the interface with ```insertComponentMode``` variable. The code is hard to read because nodeAuthoringController.ts and node.html also contain other AT features like copy/delete/edit paths/rubrics.

Move the code and templates to a new route to improve readability of the code.

hirokiterashima:
Extracted from nodeAuthoringController and node.html into its own route and components. Cleaned up related code along the way.

Test the import component feature and make sure that everything works as before including the following:
- import from my projects and library projects
- canceling import
- importing single and multiple components
- 'componentImported' event is saved in the events table in the db

Closes #2598

### DIFF
--- a/src/main/webapp/site/src/app/hybrid-module.ts
+++ b/src/main/webapp/site/src/app/hybrid-module.ts
@@ -199,7 +199,6 @@ export class PreviewAngularJSModule {
     TeacherWebSocketService
   ],
   entryComponents: [
-    ManageStudentsComponent,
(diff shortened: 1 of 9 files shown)


geoffreykwan (approved):
Looks good.

PR #2604, opened 2020-09-04
Issue 2581 replace logout event with observables

### LINKED ISSUE #2581: Replace 'logOut' event broadcasts with Observables

ekhumalo:
Replaced logOut events with observables.

To Test:
Open the Teacher and Student Project Environments and test if `forcelogOut` in SessionService is called after
`server.servlet.session.timeout=30m` in application.properties. You can reduce temporarily change time to 1 minute for testing purposes

Closes #2581


hirokiterashima (changes requested):
It looks like messages.xlf has been wiped clean. Can you revert it? Did you have the pre-commit git hook set up properly?

hirokiterashima (approved):
Looks good.

PR #2607, opened 2020-09-09
Extracted add component feature into its own route.

### LINKED ISSUE #2602: Refactor: extract create component feature into its own route
The "add new component" feature in the authoring tool is in nodeAuthoringController.ts and node.html and turned on/off in the interface with addComponentMode and showCreateComponent variables. The code is hard to read because nodeAuthoringController.ts and node.html also contain other AT features like copy/delete/edit paths/rubrics.

Move the code and templates to a new route to improve the readability of the code.

hirokiterashima:
Split add component UX to two parts: one to select component type and one to choose where it goes. 

Test that you can add a new component to a step. Test canceling out of the flow as well as choosing different places to insert the new component.

Closes #2602

### DIFF
--- a/src/main/webapp/wise5/authoringTool/addComponent/addComponentModule.ts
+++ b/src/main/webapp/wise5/authoringTool/addComponent/addComponentModule.ts
@@ -1,0 +1,26 @@
+import * as angular from 'angular';
+import { ChooseNewComponent } from './choose-new-component.component';
+import { ChooseNewComponentLocation } from './choose-new-component-location.component';
+
+export default angular.module('addComponentModule', ['ui.router'])
+  .component('chooseNewComponent', ChooseNewComponent)
+  .component('chooseNewComponentLocation', ChooseNewComponentLocation)
+  .config(['$stateProvider', $stateProvider => {
+    $stateProvider
+      .state('root.at.project.node.add-component', {
+        url: '/add-component',
+        abstract: true,
+        resolve: {}
+      })
+      .state('root.at.project.node.add-component.choose-component', {
+        url: '/choose-component',
+        component: 'chooseNewComponent'
+      })
+      .state('root.at.project.node.add-component.choose-location', {
+        url: '/choose-location',
+        component: 'chooseNewComponentLocation',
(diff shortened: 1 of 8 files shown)


geoffreykwan (changes requested):
Functionality works well.

1. I think the "Next" button should be on the right side instead of the left side.
1. It would be nice if in the "choose-location" view you could go back to the "choose-component" view.
1. When you go into the add component view, the top and side nav bars disappear which is a little jarring because it shifts the whole view. Would it be better to keep the nav bars visible?

hirokiterashima:
1- will do.
2- a change like that assumes that the user is usually adding more than one component at a time to a step, and once they're done, they'd have to "cancel" out of the add component mode. Is this how most users add components? If not, it could be annoying for users to cancel out after adding one component. Is there maybe a better way to implement the add component flow? We might want to discuss this before moving forward with changes. 
3- we do the same thing now with the structures authoring. I think we have it like that because the user is in a "structure authoring mode" where they can't do anything else but go forward or cancel, kind of like a floating dialog/wizard, and the nav menus should not be available in that mode?

@breity, what do you think?

geoffreykwan:
2. I meant adding a "Back" button in the "choose-location" view in case they realized they chose the wrong component type. It's annoying if they choose the wrong component type and get to the choose location view and then have to cancel and start the whole process from the beginning.

3. If they're restricted to the "structure authoring mode" functionality that's fine. I just think having the whole UI shift up and down is not ideal.

hirokiterashima:
2. ok, thanks for the clarification. A back button would be easy to add. Let's discuss UI/UX for add component feature in the future.
3. ok, will keep as is. Maybe there's a better way to do this in the future so it's not as jarring.

hirokiterashima:
I added the back button. The flow should remember the component type choice that you made and highlight it again when you get back to the chooser page.

I also put the back/cancel/next buttons at the bottom of the page and changed the instructions to h5. These styles better match our structured authoring.

geoffreykwan (approved):
Looks good.

PR #2617, opened 2020-09-16
Replace currentPeriodChanged event with observables

### LINKED ISSUE #2616: Replace currentPeriodChanged event with observables
### Description

Replace all the $broadcast('currentPeriodChanged') and $on('currentPeriodChanged') calls with an observable.

geoffreykwan:
Make sure changing period still works in the Authoring Tool. Test the different places where period can be changed such as
- Grade By Step
- Grade By Team
- Milestones

Closes #2616


hirokiterashima (changes requested):
The line ```this.$scope.$on('$destroy', ()``` is throwing errors like below in several files:

```
TypeError: Cannot read property '$on' of undefined
    at new WorkgroupSelectController (workgroupSelect.ts:41)
```

The constructors of those files should have ```private $scope: any``` instead of just ```$scope``` and then you can remove the line ```this.$scope = $scope;```

hirokiterashima (approved):
Looks good.

PR #2619, opened 2020-09-16
Replace currentNodeChanged and currentStepChanged events with an observable

### LINKED ISSUE #2618: Replace currentNodeChanged and currentStepChanged events with an observable
### Description

Replace all the $broadcast('currentNodeChanged') and $on('currentNodeChanged') calls with an observable.

Remove the currentStepChanged event since it is being fired but nothing is listening for it.

geoffreykwan:
Make sure all the places that listen for currentNodeChanged still works.

Closes #2618


hirokiterashima (approved):
Looks good.

PR #2629, opened 2020-09-18
JS error after wise link/button click

### LINKED ISSUE #2628: JS error after wise link/button click
### Description

When I preview the unit and click on the wise link or wise button, it takes me to the step, but I see an error in the js console:
```
TypeError: Cannot read property 'unsubscribe' of undefined
    at WiselinkController.unsubscribeAll (wiselink.js:21)
    at WiselinkController.ngOnDestroy (wiselink.js:17)
    at wiselink.js:12
    at Scope.$broadcast (angular.js:19723)
    at Scope.$destroy (angular.js:19337)
    at cleanupLastView (viewDirective.js:217)
```

It does not seem to affect behavior.

### Steps to Reproduce

1. preview a step with a html component with a wise link or button to another step.
1. click on the wise link or button
1. open js console in the developer tools

geoffreykwan:
1. Add two WISE links to an HTML component
1. Preview the project and click on one of the links
1. You should not see the console error anymore

This error occurs when there are multiple WISE links on a component. The link that was not clicked would throw the error because it would have a null currentNodeChangedSubscription. I also noticed that the function in the subscribe that calls scrollAndHighlightComponent() doesn't even get called. I'll create another issue for that.

Closes #2628

### DIFF
--- a/src/main/webapp/wise5/directives/wiselink/wiselink.js
+++ b/src/main/webapp/wise5/directives/wiselink/wiselink.js
@@ -18,7 +18,9 @@ class WiselinkController {
   }
 
   unsubscribeAll() {
-    this.currentNodeChangedSubscription.unsubscribe();
+    if (this.currentNodeChangedSubscription != null) {
+      this.currentNodeChangedSubscription.unsubscribe();
+    }
   }
 
   $onInit() {


hirokiterashima (approved):
Verified fix.

PR #2667, opened 2020-09-28
Replace notAllowedToEditThisProject and notLoggedInProjectNotSaved events

### LINKED ISSUE #2665: Replace notAllowedToEditThisProject and notLoggedInProjectNotSaved events
### Description

Replace these events with observables.

geoffreykwan:
Make sure the appropriate message is displayed in the upper right of the Authoring Tool when the project is saved but receives an error response.

You can set a breakpoint in handleSaveProjectResponse() and then manually set these values

To test notAllowedToEditThisProject
response.status = 'error'
response.messageCode = 'notAllowedToEditThisProject'

To Test notLoggedInProjectNotSaved
response.status = 'error'
response.messageCode = 'notSignedIn'

Closes #2665


src/main/webapp/wise5/services/projectService.ts
```diff
@@ -38,6 +38,12 @@ export class ProjectService {
   transitions: any;
   private errorSavingProjectSource: Subject<any> = new Subject<any>();
   public errorSavingProject$: Observable<any> = this.errorSavingProjectSource.asObservable();
+  private notAllowedToEditThisProjectSource: Subject<any> = new Subject<any>();
+  public notAllowedToEditThisProject$: Observable<any> =
+      this.notAllowedToEditThisProjectSource.asObservable();
+  private notLoggedInProjectNotSavedSource: Subject<any> = new Subject<any>();
+  public notLoggedInProjectNotSaved$: Observable<any> =
+      this.notLoggedInProjectNotSavedSource.asObservable();
```
hirokiterashima, line 46: Should these be in TeacherProjectService instead?

hirokiterashima (dismissed):

src/main/webapp/wise5/services/projectService.ts
geoffreykwan: There seems to be a good amount of code in the ProjectService that is only used by teachers, such as saving the project. Should we make another issue to move all the teacher code from ProjectService to TeacherProjectService?

hirokiterashima: Sounds good. I'll create a separate issue.

hirokiterashima (approved):

PR #2710, opened 2020-10-13
Converted add component views to Angular

### LINKED ISSUE #2709: Convert add component views to Angular component
### Description

Convert "choose new component" and "choose location" views from AngularJS component to Angular component. Then use downgradeComponent so that it will work with the AngularJS Authoring Tool.

hirokiterashima:
Converted choose new component and choose location views to Angular, then downgraded them so they work with AngularJS Authoring Tool.

Test that the add component feature works as before. Test that you can tab through the component choices using "Tab" and "Enter" keys.

Closes #2709


geoffreykwan (approved):
Looks good.

PR #2713, opened 2020-10-15
Converted import step views to Angular components

### LINKED ISSUE #2712: Convert import step views to Angular components
### Description

Convert "choose step to import" and "choose imported step location" views from AngularJS component to Angular component. Then use downgradeComponent so that it will work with the AngularJS Authoring Tool.

hirokiterashima:
Converted "choose step(s) to import" and "choose location for imported steps" views to Angular, then downgraded them so they work with AngularJS Authoring Tool.

Test that the import step feature works as before.
- import from my/library unit
- import one/multiple steps
- import as the first step and after an existing step
- cancel out of the feature

Closes #2712


geoffreykwan (approved):
Looks good.

Issue #2716, opened 2020-10-16
Fix serverSaveTime type inconsistencies


hirokiterashima:
ServerSaveTime sent over websocket is a number
ServerSaveTime sent via GET request is a timestamp string

This inconsistency was introduced when we added JSONSerialization in this commit 868960d8931ba49eb69225327af601521a85bfaa

In this case, we should sent a number representing the time.

PR #2717, opened 2020-10-16
Converted ComponentNewWorkBadge to Angular

### LINKED ISSUE #2715: Upgrade ComponentNewWorkBadge to Angular Component
### Description

Upgrade the component to Angular, and downgrade to AngularJS so that it could be used in the AngularJS classroom monitor.

hirokiterashima:
Test that the "New" badge shows up when a new student work comes in while the teacher is grading.

Right now, due to bug #2716, the "New" badge won't show properly if you open the classroom monitor after the new work has been submitted. It'll only show up if the work was sent over the websocket, and the classroom monitor is already open.

Closes #2715


geoffreykwan (approved):
Looks good.

PR #2721, opened 2020-10-19
DrawAuthoringController now extends ComponentAuthoringController

### LINKED ISSUE #2720: DrawAuthoringController extend ComponentAuthoringController
Change DrawAuthoringController to extend from ComponentAuthoringController. It currently extends DrawController.

Change the way we save starter drawing to use observables.

hirokiterashima:
Test that draw component authoring works as before. Starter drawing are now sent as events using Observables, which was required to break the dependency between the authoring and viewing controllers.

Specifically:
- saving starter drawing
- connected components with various component types

Closes #2720


geoffreykwan (approved):
Looks good.

Issue #2723, opened 2020-10-19
Step advanced view fails to load in prod build


hirokiterashima:
### Description
The step advanced view doesn't load in build-prod version of the application. build-dev version works. This is likely related to disambiguation of variable names. An error is thrown in the console:
```
angular.js:15570 Error: [$injector:unpr] Unknown provider: tProvider <- t
https://errors.angularjs.org/1.7.9/$injector/unpr?p0=tProvider%20%3C-%20t
    at angular.js:138
```

### Steps to Reproduce

1. npm run build-prod
2. open step advanced authoring

PR #2724, opened 2020-10-19
Fixed issue where step advanced view was not loading

hirokiterashima:
The issue was the optimization phase of the production build was shortening the variables names, and the $injector could not find the right dependencies. I fixed this by adding $inject to the classes.

Test that node advanced authoring loads.

#2723


geoffreykwan (approved):
Looks good.

PR #2728, opened 2020-10-19
Re-enable custom unit styles

### LINKED ISSUE #2727: Restore support for custom unit styles
ProjectService has a has a `getStyle()` function that returns the `style` key from `project.json`. We used to use this to allow authors to add custom styles per unit.

We don't actually do anything with these custom styles anymore. We should re-enable.

breity:
Re-enabled support for custom unit styles.

To test:
- Add a "style" key and value to a unit's `project.json` file with custom CSS. For example:
```
"style": "p {color: red;}"
```
- Preview the unit and check that your custom styles are applied.

Closes #2727.


hirokiterashima:
This didn't work for me. I added the style key/value to project.json but none of the text showed up red...it looks like it should work. Am I missing something?

hirokiterashima (approved):

PR #2730, opened 2020-10-20
Moved SnipImage subscription code to ComponentController

### LINKED ISSUE #2729: Move SnipImage code from component to componentController
### Description

Snip image is only available for students and preview users (and not authors/teachers), so the code should be moved to componentController

hirokiterashima:
Test that you can still snip image in preview and student modes.

Closes #2729


geoffreykwan (approved):
Looks good.

PR #2731, opened 2020-10-20
Fixed server and client timestamp inconsistencies

### LINKED ISSUE #2716: Fix serverSaveTime type inconsistencies

hirokiterashima:
Changed server and client save times returned from the server to a long instead of a timestamp for StudentWork, Annotation, and Event.

Test that the "new" work badge appears correctly in the grading tool when student has ungraded work, either retrieved from the GET request (on grading tool load), or from WebSocket (student submits work while teacher is looking at the component in the grading tool).

Closes #2716


geoffreykwan (approved):
Looks good.

PR #2734, opened 2020-10-20
Graph can't set plot type to scatter or column

### LINKED ISSUE #2719: Graph can't set plot type to scatter or column
### Description

1. Go to a step that contains a Graph component or create one
1. Change the "Plot Type" to "Scatter Plot"
1. Preview the step
1. Draw two points
1. A line plot will be drawn but it should be a scatter plot. Column plot does not work either.

geoffreykwan:
Things to test
- Change the graph plot type and make sure all the series types also change to match that type.
- Change the graph plot type to "Column Plot" and make sure the X Axis type is changed to "Categories".
- If the graph plot type is "Line Plot" or "Scatter Plot", make sure you can add a new series and change the series type to a different type. This is so you can have a line series and a plot series on the same graph.

Closes #2719


hirokiterashima (approved):
Looks good.

PR #2737, opened 2020-10-20
Extracted EditComponent from component.js

### LINKED ISSUE #2736: Extract EditComponent from Component
Right now, component.js contains code for both editing and viewing the component. Extract the editing component code out of the file and into its own component. This will make the code simpler and easier to upgrade to Angular component.

hirokiterashima:
Also cleaned up code. editComponent no longer needs to call $compile.

Test that editing components in the AT works as before, and also test student/grading views to make sure that they still work.

Closes #2736


geoffreykwan (commented):
Looks good.

PR #2742, opened 2020-10-22
Extracted preview component from component.js

### LINKED ISSUE #2741: Extract PreviewComponent from Component
Right now, component.js contains code for both previewing the component and viewing the component as a student. Extract the previewing component code out of the file and into its own component. This will make the code simpler and easier to upgrade to Angular component.

hirokiterashima:
Code to handle previewing a component in the authoring tool has been extracted from component.js into its own component. Also cleaned up code in component.js. Note that component.js no longer uses $compile().

Test that previewing component works in the AT. Also test that component works as before in student and grading views.

Closes #2741


geoffreykwan (approved):
Looks good.

PR #2751, opened 2020-10-24
Fixed issue where inactive tab + active tab was still logging users out

### LINKED ISSUE #2750: Don't logout if another tab is active
### Description

When you are working on WISE with two tabs open
- one tab is active
- the other is inactive

The inactive tab should not cause your session to be inactivated

hirokiterashima:
Open two tabs, and open a unit in the AT in each tab. You should reduce your session timeout value in application.properties.

Test that
- being inactive in one and active in another does not prevent either tab from logging out
- logging out of one tab will log you out of another

Also fixed issue where users were not being logged out when the session was inactive. It was testing for string 'false', when the server was returning boolean false.

Closes #2750


breity (approved):
Looks good. Side effect of this change is that session automatically renews every minute even in inactive tabs now. So user will not be logged out due to inactivity unless all tabs are closed. Will approve and merge this and create new issue to reinstate automatic logout due to inactivity.

PR #2767, opened 2020-10-27
Fixed issue where assigning scores and comments to components was cau…

### LINKED ISSUE #2766: Bug giving score/comment causes glitch in grading tool
Report from a WISE Teacher:
In "Grading by Step", every time I give a student a score (or comment), it changes from that step being "Completed" to "Not Assigned" to the student. I cannot reopen (expand) that students answer to change a score or give a comment without refreshing the page every time. I can give a comment or score but then have to refresh and go back to that student to change or add anything - I can not score and then comment at once like I used to be able to do.

hirokiterashima:
This fixes the issues where assigning scores and comments to components caused it to show "Not Assigned" and prevented the teacher from further grading.

The issue had to do with inconsistency in the way we handle toWorkgroupId and fromWorkgroupId. Sometimes it is a string, and sometimes it is a number. This caused the annotation lookup (this.annotations[toWorkgroupId_string]) to return null, even when it does in fact exist in the array.

This is a temporary hot fix. The long term fix should be to always pass in integers from parent component to the child component. I'll create a separate issue for that.

Closes #2766

### DIFF
--- a/src/main/webapp/wise5/directives/component/component.ts
+++ b/src/main/webapp/wise5/directives/component/component.ts
@@ -11,8 +11,8 @@ class ComponentController {
   componentState: any;
   mode: string;
   nodeId: string;
-  teacherWorkgroupId: number;
-  workgroupId: number;
+  teacherWorkgroupId: string;
+  workgroupId: string;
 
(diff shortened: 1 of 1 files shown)


geoffreykwan (approved):
Looks good.

PR #2770, opened 2020-10-27
Remove nodeAuthoringController and $scope from editComponentController

### LINKED ISSUE #2769: Remove nodeAuthoringController dependency from editComponentController
- Use observables instead to communicate between the two controllers. 
- Remove $scope from editComponentController.

These are preventing us from upgrading this component to Angular.

hirokiterashima:
Removing these dependencies will allow us to upgrade this component to Angular.

Instead of doing $scope.$watch for changes to authoringComponentContent, the component authoring components should call authoringViewComponentChanged(). This was already being called from most places, so I figured it was safe to remove the $scope.$watch block.

Test that editing components works as before, especially editing advanced component settings.

Closes #2769


geoffreykwan (approved):
Looks good.

PR #2772, opened 2020-10-28
Changed workgroupId and teacherWorkgroupId inputs to be numbers

### LINKED ISSUE #2771: Pass in numbers for workgroupId and teacherWorkgroupId to component directive
We have an inconsistency in the way we handle toWorkgroupId and fromWorkgroupId. Sometimes it is a string, and sometimes it is a number. This causes the annotation lookup (this.annotations[toWorkgroupId_string]) to return null, even when it does in fact exist in the array. We created a hotfix in #2767.

Fix this for good. The workgroupId and teacherWorkgroupId in component should always be a number.

hirokiterashima:
These inputs to the component directive will always be a number. 

This fixes the inconsistency in the way we handled toWorkgroupId and fromWorkgroupId. Sometimes it was a string, and sometimes it was a number. This caused the annotation lookup (this.annotations[toWorkgroupId_string]) to return null, even when it does in fact exist in the array. We created a hotfix in #2767. This PR fixes it for good.

Test that grading, viewing revisions, and working on the component as a student work as before.

Closes #2771


geoffreykwan (approved):
Looks good.

PR #2778, opened 2020-10-31
Removed OnlyShowWork mode and removed $rootScope from NodeController

### LINKED ISSUE #2777: Code Clean Up: NodeController, remove 'onlyShowWork' mode
- Remove 'onlyShowWork' mode and related code in component controllers
- Remove $rootScope from NodeController
- Instead of writing ```this.$filter = $filter;```, decorate the constructor argument ```protected $filter: any,```

hirokiterashima:
OnlyShowWork mode in the component directive was not being used, so I removed it.

I also cleaned up related code. For example, I replaced ```this.$filter = $filter``` line in the constructor with ```private $filter: any``` parameter argument to the constructor.

Test that student vle and grading tools work as before.

Closes #2777

### DIFF
--- a/src/main/webapp/wise5/components/componentController.ts
+++ b/src/main/webapp/wise5/components/componentController.ts
@@ -15,23 +15,7 @@ import { Directive } from '@angular/core';
 
 @Directive()
 class ComponentController {
-  $filter: any;
-  $injector: any;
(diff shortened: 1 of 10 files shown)


geoffreykwan (approved):
Looks good.

PR #2784, opened 2020-11-02
Upgraded NodeAdvancedJSONAuthoring component to Angular

### LINKED ISSUE #2782: Upgrade NodeAdvancedJSONAuthoring component to Angular
Upgrade the component to Angular, and downgrade to AngularJS so that it could be used in the AngularJS authoring tool.

hirokiterashima:
Test that you can make changes to the node JSON in the advanced node authoring view. It should debounce every second like before. Unfortunately Angular does not have a debounce attribute so I had to write an observer.

Closes #2782


geoffreykwan (approved):
Looks good.

PR #2789, opened 2020-11-03
Separated hybrid angular bundle into teacher and student

### LINKED ISSUE #2788: Separate hybrid bundle into teacher and student bundles
Right now the build wraps both teacher and student files into one hybrid-module bundle. 

There are files that only teachers use (like grading and authoring files), and files that only students use (like nodeController, vleController).

Change the build to spit out three bundle files: common, teacher, and student. Teachers should load common+teacher bundles, and students should load common+student bundles.

This will improve the load time for all users.

hirokiterashima:
This PR separates the hybrid angular js apps' bundles into their own bundles, teacher-hybrid-angular-module and student-hybrid-angular-module, plus a common bundle. Before, there was just one hybrid-module that both teachers and students loaded, and this was forcing them to download files that they would never use.

Test that everything works in both development build and production build.

Another benefit (besides faster loading for users) is faster build time for developers. With the separation, Angular doesn't need to build the entire hybrid-module app bundle, but instead only build the bundle for the app (teacher, student, or common) that you changed.

Closes #2788


breity (approved):
Great, looks good.

geoffreykwan (approved):
Looks good.

PR #2795, opened 2020-11-05
Upgraded NodeStatusIcon to Angular

### LINKED ISSUE #2794: Upgrade NodeStatusIcon to Angular Component
Upgrade the component to Angular, and downgrade to AngularJS so that it could be used in the AngularJS pages.

hirokiterashima:
I removed the directive from several places that weren't using them, like NodeAdvanced[Constraint/Branch/Path]Authoring template files. They appear to be a byproduct of copy&paste.

Test that the completion status icon appears next to the step drop down and in the unit plan view when you log in as a student and complete a step.

Closes #2794


geoffreykwan (approved):
Looks good.

PR #2799, opened 2020-11-06
Upgraded WorkgroupInfo to Angular component

### LINKED ISSUE #2797: Upgrade WorkgroupInfo to Angular Component
Upgrade the component to Angular, and downgrade to AngularJS so that it can be used in the AngularJS pages.

hirokiterashima:
Also fixed an issue where nodeStatusIcon aria-label was preventing messages extraction from working.

Test that the workgroup-info component works correctly. In grade-by-step view, it should show the account avatar icon, team members' name, and the team id. If there is a notification (e.g. ambient notification based on c-rater), it should appear as well, like below.
Screen Shot 2020-11-06 at 11 48 40 AM

Closes #2797


geoffreykwan (changes requested):
The new work icon doesn't seem to show up in the grade by step view that lists out all the workgroups. On PROD it doesn't seem to work either though.

geoffreykwan (approved):
Looks good. A new issue has been created for the new work icon bug #2805.

PR #2800, opened 2020-11-06
Disabled run and unit sharing from admin pages

### LINKED ISSUE #2798: Temporarily disable run and unit sharing from admin tools
Since we updated the permissions for runs and units, sharing from the old admin and researcher tools doesn't work properly. We should disable sharing from there until the admin section is rewritten to work with the new permissions.

Locations where the link to manage shared teachers should be removed:
- `/admin/run/stats`

image

- `/admin/run/manageprojectruns.html`

image

- `/admin/run/stats-by-activity`

image

- `/admin/project/manageallprojects.html`

image

image

- We should also put back the unit (project) management section for researcher accounts so that they can utilize the project search tool. It's currently hidden for researcher accounts but visible to admins.

image

hirokiterashima:
Also show Project Management section to researchers.

Test the following:
- run/unit sharing options do not appear in the admin pages.
- researchers (not admins) can view and use the Project Management section  in the admin home page.

Closes #2798

### DIFF
--- a/src/main/webapp/portal/admin/index.jsp
+++ b/src/main/webapp/portal/admin/index.jsp
@@ -172,33 +172,31 @@
 
                     </div>
 
-                    <sec:authorize access="hasRole('ROLE_ADMINISTRATOR')">
-                        <div class="sectionHead">
-                            <spring:message code='admin.index.projectManagement' />
-                        </div>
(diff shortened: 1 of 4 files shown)


breity (approved):
Looks good. 👍

PR #2804, opened 2020-11-09
Add ETS logo to site footer

### LINKED ISSUE #2803: Add ETS logo to site footer
We should add the ETS logo to the site footer since they're a main collaborator.

breity:
Added ETS logo to site footer.
Also fixed an issue that was preventing Angular i18n extraction to complete: `ERROR in Can't bind to 'aria-label' since it isn't a known property of 'mat-icon'`.

To test:
- Verify that ETS logo (with link) is included in site footer.
- Verify that "Back to top" button is hidden when scrolled all the way to bottom of page but reappears when scrolling up. (This is so footer content is not obscured by button on smaller screens.)

Closes #2803.


hirokiterashima (approved):
Looks good.

PR #2810, opened 2020-11-16
Replace Summernote with TinyMCE in html authoring and rubric authoring

### LINKED ISSUE #2809: Replace Summernote with TinyMCE in html authoring and rubric authoring
### Description

- Replace the rubric authoring editor in the project, step, and component rubrics
- Replace the html authoring editor in the HTML component
- Make sure Insert WISE Asset still works
- Make sure Insert WISE Link still works

geoffreykwan:
- Make sure Insert WISE Asset still works (note that locally referenced videos do not show a preview or play in TinyMCE, I'm not sure  how easy it is to fix that)
- Make sure Insert WISE Link still works (note that WISE Links don't work in the rubric and never have. I think it doesn't work because the rubric content isn't compiled like the html content)
- Make sure rubric authoring works for the project, step, and component
- Make sure HTML component authoring works
- See if we want to add or remove some TinyMCE plugins
  - Here's a [demo page](https://www.tiny.cloud/docs/demo/full-featured/)
  - Here's the [plugin page](https://www.tiny.cloud/apps/)
- I noticed that some styling for tables doesn't work properly. If you right click > Table Properties and set "Cell spacing" or "Cell padding", it will look correct in TinyMCE but it won't render correctly when previewed in WISE. I think this is a bug in TinyMCE as referenced [here](https://github.com/tinymce/tinymce/issues/3619).

Closes #2809


geoffreykwan:
This pull request has been replaced with #2818

PR #2812, opened 2020-11-17
Upgraded ComponentAnnotations to Angular

### LINKED ISSUE #2811: Upgrade ComponentAnnotations to Angular component
Upgrade the component to Angular, and downgrade to AngularJS so that it can be used in the AngularJS pages.

hirokiterashima:
I moved the component to student module (before it was in common module) because only students use this component.

Test the following:
- Show only score
- Show only comment
- Show both score and comment
- "New" label shows up for new annotations
- "Computer Feedback" title shows up for C-Rater feedback
- Hovering over timestamp should show tooltip.

Closes #2811


geoffreykwan (changes requested):
Functionality works well.

The label for "Teacher Feedback" and "Computer Feedback" is really big though. It's the biggest text on the page. It looks like with the new mat-card-title element, the mat-card-title class is automatically applied which uses font-size: 24px and overrides our annotations styling which uses font-size: 15px. Should we make it render with font-size: 15px like it used to?

hirokiterashima:
Good catch. I fixed it.

geoffreykwan (approved):
Looks good.

PR #2814, opened 2020-11-17
Changed StudentAssetController to return StudentAsset object

### LINKED ISSUE #2813: Use Serializer to serialize StudentAsset
Right now StudentAsset has a ```toJSON()``` method that creates a JSON containing fields to send to the client as a JSON string. This should instead be a serialized StudentAsset object.

Implement a JsonSerializer to serialize the StudentAsset object and change the methods in StudentAssetController to return a StudentAsset object instead of a String.

This will make the code cleaner and easier to read.

hirokiterashima:
Implemented a Serializer that serializes the StudentAsset object and changed the controller methods to return it instead of String'ified json. Also cleaned up client code that handles the response from the server.

Test that retrieving, copying, and deleting StudentAssets work as before.

As a student:
- should retrieve StudentAssets by loading a unit
- should create new StudentAsset by:
   - creating an image note
   - uploading image to discussion
   - importing concept map as background to draw
- should copy/delete StudentAsset by:
   - importing OR to label

Closes #2813

### DIFF
--- a/src/main/java/org/wise/portal/service/vle/wise5/VLEService.java
+++ b/src/main/java/org/wise/portal/service/vle/wise5/VLEService.java
@@ -122,12 +122,7 @@ public interface VLEService {
       String localNotebookItemId, Integer notebookItemId, String type, String data,
       String clientSaveTime) throws ObjectNotFoundException;
 
-  /**
-   * @return StudentsAssets from data store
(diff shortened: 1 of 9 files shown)


geoffreykwan (approved):
Looks good.

PR #2816, opened 2020-11-19
Added Serializer for Achievement

### LINKED ISSUE #2815: Use Serializer to serialize Achievement and create AchievementController
Right now Achievement has a toJSON() method that creates a JSON containing fields to send to the client as a JSON string. This should instead be a serialized Achievement object.

Implement a JsonSerializer to serialize the Achievement object and move Achievement-related methods in StudentDataController to AchievementController and return an Achievement object instead of a String.

This will make the code cleaner and easier to read.

hirokiterashima:
Also extracted Achievement-related methods from StudentDataController into AchievementController.

I left Achievement.toJSON() method because it is still being used to send WebSocket payload in broadcastAchievementToTeacher method. The way we handle WebSocket payload is a bit different from HTTP payload. This should be refactored separately later.

Test that saving and retrieving Achievements work as before in these scenarios:
- loading student vle should retrieve student's achievements.
- loading teacher grading tool should retrieve all the achievements for a run.
- completing a milestone activity (e.g. CRater) should create a new achievement. It should also send that Achievement to the teacher's grading tool in real-time and update the Milestone view.

Closes #2815

### DIFF
--- a/src/main/java/org/wise/portal/service/work/AchievementJsonModule.java
+++ b/src/main/java/org/wise/portal/service/work/AchievementJsonModule.java
@@ -1,0 +1,21 @@
+package org.wise.portal.service.work;
+
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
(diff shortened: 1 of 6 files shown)


geoffreykwan (approved):
Looks good.

PR #2825, opened 2020-12-04
Extracted component json and rubric authoring to Angular components

### LINKED ISSUE #2824: Extract rubric and json component authoring and upgrade to Angular component
Extract the component rubric and json authoring features and make them into Angular Components. Then, downgrade them so they can be used in AngularJS

hirokiterashima:
I made the component json authoring into edit-component-json directive and the component rubric authoring into edit-component-rubric directive. This reduced a lot of duplications in the template files.

Test that those features work as before.

Closes #2824


geoffreykwan (changes requested):
1. Opening the advanced view saves a snapshot of the component JSON in the edit component JSON component. This means if you open the advanced view and then change anything in the component and then go to edit the JSON, the JSON will be old. This is not reproducible on PROD. Steps to reproduce
  i. Open the advanced view for a component
  ii. Change the text in the component prompt (using the regular prompt authoring textarea)
  iii. Open the JSON view
  iv. The JSON will show the old prompt from before your recent change

2. Edit the component JSON and save it by clicking the JSON button. Now in order to open the Advanced component view for that component, you need to click the advanced button twice but it should only require one click. This is reproducible on PROD.

hirokiterashima:
1. Good catch. Fixed.
2. Let's leave this behavior for now. I think this occurs because when the json view is closed, it changes the component content and forces a recompile&re-render. It was introduced when we refactored the author and preview views. This should be fixed in the future when we move away from this recompile/re-render AngularJS.

geoffreykwan:
Should we be subscribing to nodeChanged$ or componentChanged$? Right now we're subscribing to nodeChanged$ which means the JSON gets updated for a component even if a different component in the step is changed.

hirokiterashima:
I think ultimately we do. The issue is that many parts of the AT still calls nodeChanged (componentChanged was just introduced in this PR), and changing all of those places could potentially introduce new issues. I think we should handle nodeChanged -> componentChanged conversions in a new issue.

For the time being, we can add a call to componentChanged() every time nodeChanged() is called 
```
  nodeChanged(doParseProject: boolean = false): void {
    this.nodeChangedSource.next(doParseProject);
    this.componentChanged();   // <- add this line
  }
```
...and then in edit-component-json.component.ts, change the subscription to componentChanged(). This will ensure that things will continue to work, and start moving us in that direction.

What do you think?

geoffreykwan (approved):
Looks good.

PR #2847, opened 2020-12-15
Upgraded ComponentSelect to Angular

### LINKED ISSUE #2846: Upgrade ComponentSelect to Angular
Upgrade the component to Angular, and then downgrade it to AngularJS so that it can be used in the AngularJS pages.

hirokiterashima:
Test that component select drop-down in the grading tool works as before. Selecting/unselecting items should show/hide the components.

Closes #2846


geoffreykwan (approved):
Looks good.

PR #2858, opened 2021-01-05
Constraint without Removal Criteria Name breaks preview project

### LINKED ISSUE #2856: Constraint without Removal Criteria Name breaks preview project
### Description

2. Open a step
3. Add a constraint and leave the "Removal Criteria Name" as "Please Choose a Removal Criteria"
4. Preview the project and it won't load and will just show a blank screen. You will see this error in the console

```
TypeError: this.criteriaFunctionNameToFunction[criteria.name] is not a function
    at StudentDataService.evaluateCriteria (studentDataService.ts:463)
    at StudentDataService.evaluateMultipleRemovalCriteria (studentDataService.ts:437)
```

geoffreykwan:
2. Open a step
3. Add a constraint to the step but do not set the "Removal Criteria Name", leave it as "Please Choose a Removal Criteria"
4. Preview the project and make sure the project loads (it used to throw an error in the console and then just show a blank screen)

Closes #2856

### DIFF
--- a/src/main/webapp/wise5/services/studentDataService.ts
+++ b/src/main/webapp/wise5/services/studentDataService.ts
@@ -457,8 +457,12 @@ export class StudentDataService extends DataService {
     return true;
   }
 
-  evaluateCriteria(criteria) {
-    return this.criteriaFunctionNameToFunction[criteria.name](criteria);
+  evaluateCriteria(criteria: any): boolean {
+    const criteriaFunction = this.criteriaFunctionNameToFunction[criteria.name];
+    if (criteriaFunction == null) {
+      return true;
+    }
+    return criteriaFunction(criteria);
   }
 
   evaluateIsCompletedCriteria(criteria) {


src/main/webapp/wise5/services/studentDataService.ts
```diff
@@ -457,8 +457,12 @@ export class StudentDataService extends DataService {
...
   }
 
-  evaluateCriteria(criteria) {
-    return this.criteriaFunctionNameToFunction[criteria.name](criteria);
+  evaluateCriteria(criteria: any): boolean {
+    const criteriaFunction = this.criteriaFunctionNameToFunction[criteria.name];
+    if (criteriaFunction == null) {
+      return true;
+    }
+    return criteriaFunction(criteria);
```
hirokiterashima, line 465: If the issue occurs when criteria.name is empty string (""), then should we change this check to
```suggestion
    if (criteria.name === "") {
      return true;
    }
    return this.criteriaFunctionNameToFunction[criteria.name](criteria);
```

hirokiterashima (changes requested):
The fix works, but it's checking for some a null criteriaFunction when it should be checking for an empty criteria name? What do you think about my code change suggestion that specifically checks for empty criteria name instead?

hirokiterashima (approved):

PR #2864, opened 2021-01-05
Upgraded AlertStatusCorner to Angular component

### LINKED ISSUE #2859: Upgrade AlertStatusCorner to Angular
AlertStatusCorner is an alert indicator that appears at the group header level in the grading tool when there is an un-dismissed alert for that node. For example, an CRaterResult that requires the teacher to dismiss it.

Screen Shot 2021-01-05 at 10 52 08 AM
 
1. Upgrade AlertStatusCorner to Angular and downgrade it to so that it can be used in AngularJS app.
2. Remove AlertStatusIcon.ts. It is not being used.

hirokiterashima:
Also removed AlertStatusIcon because it wasn't being used.

AlertStatusIcon is displayed in the grading tool when there is a CRater item with a dismissal code. It appears at the top-right corner of a lesson that contains the item, like this:

Screen Shot 2021-01-05 at 10 52 08 AM

Test that the alert icon appears in the grading tool when you trigger the dismissal code sequence of a CRater item from the student side.

I noticed that the icon will remain even if you dismiss the notification. It will disappear when you refresh the page. This seems to be the original behavior. We can fix this in a separate issue.

Closes #2859


geoffreykwan (approved):
Looks good.

PR #2866, opened 2021-01-06
Added support to format TypeScript files with Prettier

### LINKED ISSUE #2861: Format TypeScript files with Prettier
Use Prettier and VSCode plugin to format changes to TypeScript on paste and save. Also format existing *.ts files.

hirokiterashima:
Also added npm script to batch format files in the future.

This will make sure that our TypeScript files are formatted consistently on save and paste.

Note about trying to implement [Google's indent continuation lines at least +4 spaces guideline](https://google.github.io/styleguide/jsguide.html#formatting-indent) : I could not find an easy way to add 4 spaces at the beginning of a continuation line. The default behavior is 2 spaces. I also found this issue on Prettier: https://github.com/prettier/prettier/issues/5865. It appears that Prettier by default puts the long line into separate lines. This is for function parameters though, e.g.

```
  getMultipleAttemptCRaterFeedbackTextByScore( component: any, previousScore: any, currentScore: any) {
```
becomes
```
  getMultipleAttemptCRaterFeedbackTextByScore(
    component: any,
    previousScore: any,
    currentScore: any
  ) {
```

Perhaps we can go with the default Prettier formatting for now. What do you think?

How to test:
1. download Prettier plugin for VSCode, if you haven't already.
2. make sure that VSCode's default formatter is set to "esbenp.prettier-vscode", and not set to null
Screen Shot 2021-01-06 at 10 32 44 AM
3. make sure that formatting on save and on paste are enabled in VSCode. Follow this guide: https://khalilstemmler.com/blogs/tooling/prettier/#Formatting-using-VSCode-on-save-recommended. 
4. test that the file formats when you save, or paste in a block of code into the file.

Closes #2861


geoffreykwan (approved):
Looks good.

I actually like having the long list of parameters separated onto multiple lines because it makes it easier to read and also easier to view parameter changes in diffs.

The lack of double indent on continuation lines isn't that bad because the closing paren is placed on the next line which then visually separates the parameters from the body.

```
  getMultipleAttemptCRaterFeedbackTextByScore(
    component: any,
    previousScore: any,
    currentScore: any
  ) {
    const scoringRule = {};
```

hirokiterashima:
Yah I agree with @geoffreykwan about the parameter separations. This shouldn't happen as often if we keep our params shorter (shorter param names, less param count).

The other thing is chained property accesses or chained method calls.
```
this.is.a.really.long.property.access
```
becomes
```
this.is.a.really.long
  .property.access
(two spaces)
``` 
We can live with this? It still gives us an indentation that indicates that it's a continuation of the previous line, and also  distinguish it from the statement in the next line...just not as apparent as 4 spaces.

breity (approved):
Yep, looks good.

I agree that these defaults are fine and that putting function parameters on separate lines can be helpful too.

PR #2867, opened 2021-01-06
Upgraded Spring-Boot to 2.3.2 for WISE4

### LINKED ISSUE #2714: Upgrade Spring-boot for legacy branch
### Description

Having legacy branch running alongside develop branch will not work anymore because the sessions are handled differently due to the difference in the spring-boot/spring-security versions in those branches.

Upgrade the legacy branch to use spring-boot 2.3+.

hirokiterashima:
This will fix the issue where the session in production site and the legacy site don't match and throws an error anytime you log in from one and try to access the other (e.g. log in to prod and launch a wise4 unit preview). The problem was that the spring sessions were not the same because of different spring framework versions.

I also cherry-picked MySQL8 fixes into this branch. 

Test that you can log into WISE4 via /legacy/login and launch AT, preview and the grading tool.

Closes #2714

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,7 +314,7 @@
   <parent>
     <groupId>org.springframework.boot</groupId>
     <artifactId>spring-boot-starter-parent</artifactId>
-    <version>2.2.0.RELEASE</version>
+    <version>2.3.2.RELEASE</version>
     <relativePath/>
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #2868, opened 2021-01-07
Refactored ProjectService.getAllPaths()

### LINKED ISSUE #2396: Optimize and refactor getAllPaths in projectService.ts
Function `getAllPaths` has a Cognitive Complexity of 228 (exceeds 5 allowed). Consider refactoring. Try to get it down to 200 or less.

https://codeclimate.com/github/WISE-Community/WISE/src/main/webapp/wise5/services/projectService.ts#issue_5ed939a22c17800001000991

This method is also causing slowdown in loading/parsing the unit in the AT. Below is a Chrome Dev Tool performance profiler result from adding a new node to a unit (see below to download unit) 
Screen Shot 2020-06-04 at 1 37 58 PM

**Steps to reproduce**
1. Import this unit with 2 branches based on MC choice (15 paths each) and some activities.
[312.zip](https://github.com/WISE-Community/WISE/files/4732657/312.zip)
2. Add a new step to the unit. You will see that it takes time for the AT to show the new step.

Refactor this method and other related code so that it is less complex and performs faster.

hirokiterashima:
Also refactored consolidatePaths() and getParentGroup() and added tests. 

Code complexity went down from 228 to 192 just by removing the outer-most null check. I did the same for consolidatePaths() and getParentGroup(). We can take a closer look at getAllPaths() in a future task and get the complexity down even more.

Test that the unit preview and authoring work as before. Test with various units, with and without branches.

Closes #2396

### DIFF
--- a/src/main/webapp/site/src/app/services/projectService.spec.ts
+++ b/src/main/webapp/site/src/app/services/projectService.spec.ts
@@ -1109,15 +1109,14 @@ function consolidatePaths() {
 
 function getParentGroup() {
   describe('getParentGroup()', () => {
+    beforeEach(() => {
+      service.setProject(twoStepsProjectJSON);
+    });
     it('should get the parent group of an active node', () => {
-      service.setProject(oneBranchTwoPathsProjectJSON);
-      const parentGroup = service.getParentGroup('node1');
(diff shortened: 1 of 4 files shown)


geoffreykwan (approved):
Looks good.

PR #2888, opened 2021-01-15
Upgrade Concept Map authoring to Angular component

### LINKED ISSUE #2885: Upgrade Concept Map authoring to Angular component
Upgrade the AngularJS component to Angular.

geoffreykwan:
Make sure Concept Map authoring still works.

@breity Can you also look at why the authoring elements don't stay within the right side of the component authoring container when the screen width gets too small (like less than 1100px). This problem also happens on production.

Look at the right side of this screen shot
Screenshot 2021-01-15 151324

Closes #2885


breity (approved):
Looks good.

I fixed the small screen layout bug, updated styles to better delineate nodes and links, and improved layout.

PR #2936, opened 2021-02-16
Extract grading view from Multiple Choice component

### LINKED ISSUE #2935: Extract grading view from Multiple Choice component
Create a Multiple Choice grading component and extract the grading html from the student html.

geoffreykwan:
- Make sure Multiple Choice grading and student views still work
- Make sure Multiple Choice and Open Response student revisions show up properly in the grading view along with scores and comments for the revisions (I broke the student revisions previously when I worked on extracting the Open Response grading)
- Make sure student work timestamps show up (this seems to have been broken for a while when we switched the timestamp format we get from the server)

The template code in [workgroupNodeGrading.ts](https://github.com/WISE-Community/WISE/blob/1caa0813835627dfc12e1385fb21151490b46c58/src/main/webapp/wise5/classroomMonitor/classroomMonitorComponents/shared/workgroupNodeGrading/workgroupNodeGrading.ts#L139) and [workgroupComponentRevisions.ts](https://github.com/WISE-Community/WISE/blob/1caa0813835627dfc12e1385fb21151490b46c58/src/main/webapp/wise5/classroomMonitor/classroomMonitorComponents/shared/workgroupComponentRevisions/workgroupComponentRevisions.ts#L160) are very similar so we may want to create a component for that part in the future.

Closes #2935


hirokiterashima (approved):
Looks good.

PR #2944, opened 2021-02-23
Extract grading view from Label component

### LINKED ISSUE #2943: Extract grading view from Label component
Create a Label grading component and extract the grading html from the student html.

geoffreykwan:
Make sure Label grading and student views still work properly.

Closes #2943


hirokiterashima (approved):
Looks good. Fyi, you can use javascript's built-in String.localeCompare function to compare strings so there's no need to write it.

PR #2950, opened 2021-02-25
Added interval for compiling component preview on changes.

### LINKED ISSUE #2949: Match component authoring slows down
### Description

Match component authoring slows down after typing a long sentence in the multiple feedback input fields.

### Steps to Reproduce

1. Open match component authoring
2. type long sentences for choice feedback, switching between the different feedback inputs.

**Expected behavior:**
No noticeable slowdown

**Actual behavior:**
Noticeable slowdown. Inputs are slow to focus, and letters do not appear in real-time

hirokiterashima:
This fixes the issue where components like match that have many parts slowed down noticeably when a lot of text were inputted.
Before, the preview was updating with each change. The interval acts like a debounce, so the update does not occur as frequently.

Test that previewing component works as before, and that the match component slowdown is not as noticeable. 

Fixes #2949


geoffreykwan (approved):
Looks good.

PR #2951, opened 2021-02-25
GoogleUserNotFound page now displays again.

### LINKED ISSUE #2915: Google User Not found page is blank
Try to sign in with Google with an account that doesn't exist. The googleusernotfound route flickers for a second and then you get redirected home but the view is blank.

This should show a page saying that the google user was not found, and a link to join WISE.

hirokiterashima:
The issue was the pathing in the Login routes. Also cleaned up the Component file by removing unnecessary code. 

Test:
Log in with a Google account that is not tied to any account. It should now show the the Google user not found page.

Closes #2915


geoffreykwan (approved):
Looks good.

